Issue #19, opened 2025-07-25
Avoid passing component constructor functions to view factories


nickschaap:
Web components are designed not to accept constructor arguments in order to make them compatible with the CustomElementRegistry. I'm not sure we should be violating this design decision. Ideally consumers can provide templates literals in route state declarations.

Example:
```typescript
import { html } from "lit";
import "./welcome-view.js";

const welcomeState = {
  parent: 'app',
  name: 'welcome',
  url: '/welcome',
  component: html`<welcome-view></welcome-view>`,
};
```

### How do we inject context?

One option is having the component request the injected context when it connects similar to how the main router is discovered currently or building off @lit/context (see #18). We could potentially provide some utility decorators to simplify this for consumers.

```typescript
export class WelcomeView extends LitElement {
  @routedContext()
  private _props!: UIViewInjectedProps;

  ...

}
```

Alternatively we can always have the consumer pass function which accepts the props as an argument.

Example:
```typescript
import { html } from "lit";
import "./welcome-view.js";

const welcomeState = {
  parent: 'app',
  name: 'welcome',
  url: '/welcome',
  component: (props) => html`<welcome-view .data=${props.resolves}></welcome-view>`,
};
```

PR #22, opened 2025-07-26
wip: argless constructors

simshanith:
investigation for #19


simshanith:
leaning hard the other way, sorry @nickschaap

PR #86, opened 2026-01-06
docs: homepage hero mix-blend-mode

simshanith:


simshanith:
really like it on chrome but not good on safari (desktop okayish; mobile bad)

image

image

vs chrome desktop

Screenshot 2026-01-05 at 21 09 28
Screenshot 2026-01-05 at 21 09 25

PR #147, opened 2026-07-04
feat: TransitionController — reactive controller for UI-Router transitions

simshanith:
## Summary

- Adds `TransitionController`, a zero-dependency Lit ReactiveController in `packages/lit-ui-router` that binds a host element to `@uirouter/core` transition hooks: it discovers the router from the `<ui-router>` context on `hostConnected`, registers the configured hooks (default `onSuccess`; `onBefore`/`onStart`/`onError` opt-in, with `HookMatchCriteria` support), calls `host.requestUpdate()` per matching transition, and deregisters everything on `hostDisconnected`. On (re)connect it synchronizes from the last successful transition so sticky components never render stale state.
- Eliminates all manual refresh crutches in the sample app: `App` and `Compose` use `TransitionController` (active-section visibility, draft reset from route params with draft-preservation semantics), `NavHeader`/`MessageList` use a small `StoreCommitController` for data-store commits, and `MessageTable`'s broken `hasChanged` comparator is fixed.
- README gains a "Reacting to Transitions" section.

## Test plan

- 13 new browser-mode vitest specs for the controller (discovery, explicit router, per-event registration, cancellation via `onBefore`, disconnect/reconnect lifecycle, accessors) — full suite passes in chromium/firefox/webkit
- Cypress e2e suite passes against the built docs deploy

### DIFF
--- a/packages/lit-ui-router/typedoc.json
+++ b/packages/lit-ui-router/typedoc.json
@@ -12,7 +12,14 @@
   "excludeInternal": true,
   "excludePrivate": true,
   "categorizeByGroup": false,
-  "categoryOrder": ["core", "components", "directives", "controllers", "hooks", "types"],
+  "categoryOrder": [
+    "core",
+    "components",
+    "directives",
+    "controllers",
+    "hooks",
+    "types"
+  ],
   "readme": "none",
   "githubPages": false,
   "entryFileName": "index",
--- a/packages/lit-ui-router/typedoc.json
+++ b/packages/lit-ui-router/typedoc.json
@@ -15,8 +15,8 @@
   "categoryOrder": [
     "core",
     "components",
-    "directives",
     "controllers",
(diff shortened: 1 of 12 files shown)


packages/lit-ui-router/typedoc.json
```diff
@@ -12,7 +12,14 @@
   "excludeInternal": true,
   "excludePrivate": true,
   "categorizeByGroup": false,
-  "categoryOrder": ["core", "components", "directives", "hooks", "types"],
+  "categoryOrder": [
+    "core",
+    "components",
+    "directives",
+    "controllers",
```
simshanith:
```suggestion
    "controllers",
    "directives",
```
alphabetize

simshanith: Applied in 00004cd.

PR #157, opened 2026-07-05
chore: dependency upgrades (July 2026)

simshanith:
Six-month dependency review. Strategy per request: **wide support ranges for the shipped libraries, fresher pins for apps/examples/tooling.**

## Shipped libraries — ranges widened
- `publishedPeer` catalog gains `lit: ^3.0.0` and `mobx: ^6.0.0`; `lit-ui-router-mobx` peer deps and `lit-ui-router` runtime deps now reference it. Previously these pulled from the main catalog, so every routine catalog bump silently raised the published floors (mobx peer floor literally equaled the latest release). Import audit: every lit API used is core 3.0 (LitElement, directives, ReactiveController, decorators); every mobx API exists since 6.0 (makeAutoObservable, reaction, comparer, runInAction). Dev deps still test against latest.
- `@uirouter/core` / `typedoc` peers unchanged (already wide at `^6.0.8` / `^0.28.0`). `lit-ui-router`'s runtime `@uirouter/core` dep also moved to the publishedPeer floor (`^6.1.2` → `^6.0.8`), so a consumer on an older in-range core/lit doesn't get a duplicate copy forced into their tree.

## Freshened (apps / examples / tooling)
vitest 4.1.9 · playwright 1.61 · prettier 3.9 (reformats lit `html` templates — the sample-app source diffs) · cypress 15 · wrangler 4.107 · typedoc-plugin-markdown 4.12 · vite-plugin-checker 0.14 · lodash 4.18 · CEM analyzer 0.11 · concurrently 10 · start-server-and-test 3 · in-range lockfile refresh (lit 3.3.3, typedoc 0.28.19, typescript-eslint 8.62, eslint 9.39.4, …) · security overrides: dompurify 3.4.11, undici 6.27/7.28 · examples floor to lit-ui-router ^1.3.0.

**esbuild override at 0.28.1 + docs build target raised to `safari14.1`**: esbuild ≥0.27.7 refuses to emit destructuring when any target is below safari14.1 (real JSC <14.1 array-rest bug, [compat-table#2008](https://github.com/compat-table/compat-table/pull/2008), closed as expected behavior in [esbuild#4436](https://github.com/evanw/esbuild/issues/4436)) and has no lowering transform. Rather than cap the pin, the docs VitePress target moves from its `safari14` default to `safari14.1` (drops a 2020 browser from the docs site only; sample apps already target Safari 16+). The override itself must stay while vitepress 1.x is in the tree: its vite 5.4 resolves esbuild `^0.21.3`, affected by GHSA-67mh-4wv8-2f99 (dev-server CORS, fixed 0.25). Delete the override together with the vitepress 2 upgrade.

## `@vitest/browser` patch regenerated
The old patch targeted a content-hashed bundle file (`tester-k74mgIRa.js`) so it **silently no-op'd** on 4.1.9 — install succeeded, but coverage runs exploded with 81k cascading `Unknown event: response:response:…` errors (the tester receives its own broadcasts back). Regenerated against `tester-BJtsJFpD.js` and extended the guard to also ignore `ready` (new 4.1 handshake event with the same echo problem). ⚠️ This patch will no-op again on the next vitest bump — regenerate it each time, or consider upstreaming the fix.

## CI deflake (vitest browser-mode hang — root cause found & fixed)

CI intermittently wedged until the 6h kill: a vitest 4.1.9 browser task stalled mid-run (hit 4 of 6 uncached ubuntu runs; the affected package roamed). Streamed logs caught it live: six concurrent vitest browser processes (test + test:coverage x 3 packages) all default to **API port 63315**; a loser of the "Port in use, trying another one" rebind race can end up with browser instances attached to the wrong server — the task then waits forever for instances that never connect. A second shared resource compounds it: `test` and `test:coverage` of the same package share `node_modules/.vite`, and concurrent dep re-optimization corrupts the module graph (reproduced locally: chrome collecting 0 tests).

**Fix (689415e):** each vitest invocation pins a distinct `browser.api.port` via `VITEST_BROWSER_API_PORT` and gets a per-task vite `cacheDir`. Two fully-concurrent forced local runs green with zero port-fallback messages (previously ~8 per run).

**Guardrails added along the way** (kept — they make any future hang cheap and self-diagnosing):

- `timeout-minutes: 30` on the job; `concurrency` cancel-in-progress for PR runs
- `workflow_dispatch` `force` input → `turbo run ci --force` to reproduce without cache
- CI-only `hanging-process` vitest reporter + `--log-order=stream` so a wedged task's output (incl. handle dumps) lands in the log instead of being buffered and lost

Also: `dialog.cy.js` (from #158, via merge of main) now measures centering against `documentElement.clientWidth` — cypress 15's browser renders a width-consuming scrollbar on linux, shifting the card 7.5px off the configured-viewport center.

## Deferred (each needs its own PR)
| Upgrade | Blocker |
|---|---|
| vite 8 | vitepress 1.x bundles vite 5 internally; static-copy@4 drops vite 5. Revisit when vitepress 2 (currently alpha) ships stable |
| TypeScript 6.0 | supported by typedoc/ts-eslint now, but raises the .d.ts floor for consumers — counter to the wide-support goal for shipped libs |
| ESLint 10 | needs eslint-plugin-package-json 1.x (new `@eslint/json` peer) + config migration |
| pnpm 11 | packageManager/lockfile churn |
| @types/node 26 | stays 22.x to match `engines`/.nvmrc (node 22.17.1); bump alongside a node 24 decision |

## Verification
Full `turbo run ci`: **41/41 tasks green** — unit suites across chromium/firefox/webkit (465 + 63 tests), coverage runs (155 + 21), docs build, and Cypress 15 e2e against both sample apps via wrangler (16/16 vanilla, 16/16 mobx, clean teardown under start-server-and-test 3).

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -57,7 +57,11 @@ jobs:
           Xvfb :99 -screen 0 1280x720x24 &
           echo "DISPLAY=:99" >> "$GITHUB_ENV"
       - name: Build and Test
-        run: turbo run ci --ui=stream ${{ inputs.force && '--force' || '' }}
+        # --log-order=stream: grouped logs (the CI default) only flush a
+        # task's output when it completes, so a hung task's output — e.g.
+        # vitest's hanging-process handle dump — is swallowed. Streaming
+        # prints task-prefixed lines live instead.
+        run: turbo run ci --ui=stream --log-order=stream ${{ inputs.force && '--force' || '' }}
         env:
           TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
           TURBO_API: ${{ vars.TURBO_API }}
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -14,8 +14,7 @@ on:
         type: boolean
         default: false
 
-# A new push supersedes in-flight PR runs; without this, a run wedged by the
-# vitest browser-mode exit hang keeps a runner busy until the 6h kill.
+# Cancel superseded PR runs.
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
@@ -27,9 +26,6 @@ jobs:
     # Maintainers should review fork PRs and run CI manually.
     if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
     runs-on: ubuntu-latest
-    # Uncached green runs finish in ~16 min; vitest 4.1.x browser mode
-    # occasionally prints its summary but never exits (browsers + vite stay
-    # alive), which otherwise burns the full 6h default.
     timeout-minutes: 30
     permissions:
       contents: read
@@ -57,10 +53,6 @@ jobs:
           Xvfb :99 -screen 0 1280x720x24 &
           echo "DISPLAY=:99" >> "$GITHUB_ENV"
       - name: Build and Test
-        # --log-order=stream: grouped logs (the CI default) only flush a
-        # task's output when it completes, so a hung task's output — e.g.
-        # vitest's hanging-process handle dump — is swallowed. Streaming
-        # prints task-prefixed lines live instead.
         run: turbo run ci --ui=stream --log-order=stream ${{ inputs.force && '--force' || '' }}
         env:
           TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -7,14 +7,30 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      force:
+        description: 'Bypass turbo cache (deflake reproduction: rebuild/retest everything)'
+        type: boolean
+        default: false
+
+# A new push supersedes in-flight PR runs; without this, a run wedged by the
+# vitest browser-mode exit hang keeps a runner busy until the 6h kill.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   build_and_test:
     # Security: Only run on first-party PRs (not forks) to protect secrets.
     # Fork PRs don't have access to secrets, so they would fail anyway.
     # Maintainers should review fork PRs and run CI manually.
-    if: ${{ github.event_name == 'push' || ( github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository ) }}
+    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
     runs-on: ubuntu-latest
+    # Uncached green runs finish in ~16 min; vitest 4.1.x browser mode
+    # occasionally prints its summary but never exits (browsers + vite stay
+    # alive), which otherwise burns the full 6h default.
+    timeout-minutes: 30
     permissions:
       contents: read
       pull-requests: write # Required for PR coverage comments
@@ -41,7 +57,7 @@ jobs:
           Xvfb :99 -screen 0 1280x720x24 &
           echo "DISPLAY=:99" >> "$GITHUB_ENV"
       - name: Build and Test
-        run: turbo run ci --ui=stream
+        run: turbo run ci --ui=stream ${{ inputs.force && '--force' || '' }}
         env:
           TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
           TURBO_API: ${{ vars.TURBO_API }}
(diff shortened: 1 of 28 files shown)


.github/workflows/build-test.yml
```diff
@@ -7,14 +7,30 @@ on:
...
     # Security: Only run on first-party PRs (not forks) to protect secrets.
     # Fork PRs don't have access to secrets, so they would fail anyway.
     # Maintainers should review fork PRs and run CI manually.
-    if: ${{ github.event_name == 'push' || ( github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository ) }}
+    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
     runs-on: ubuntu-latest
+    # Uncached green runs finish in ~16 min; vitest 4.1.x browser mode
+    # occasionally prints its summary but never exits (browsers + vite stay
+    # alive), which otherwise burns the full 6h default.
+    timeout-minutes: 30
```
```suggestion
    timeout-minutes: 30
```

.github/workflows/build-test.yml
```diff
@@ -41,7 +57,11 @@ jobs:
           Xvfb :99 -screen 0 1280x720x24 &
           echo "DISPLAY=:99" >> "$GITHUB_ENV"
       - name: Build and Test
-        run: turbo run ci --ui=stream
+        # --log-order=stream: grouped logs (the CI default) only flush a
+        # task's output when it completes, so a hung task's output — e.g.
+        # vitest's hanging-process handle dump — is swallowed. Streaming
+        # prints task-prefixed lines live instead.
+        run: turbo run ci --ui=stream --log-order=stream ${{ inputs.force && '--force' || '' }}
```
```suggestion
        run: turbo run ci --ui=stream --log-order=stream ${{ inputs.force && '--force' || '' }}
```

apps/sample-app-lit-e2e/src/integration/dialog.cy.js
```diff
@@ -54,10 +54,14 @@ describe('confirmation dialog', () => {
       });
     cy.get('.dialog .content').then(($content) => {
       const rect = $content[0].getBoundingClientRect();
-      const viewportCenter = Cypress.config('viewportWidth') / 2;
+      // Measure against the document's client width, not the configured
+      // viewport: a classic (non-overlay) vertical scrollbar narrows the
+      // layout area, shifting the centered card by half the scrollbar width.
+      const layoutWidth = $content[0].ownerDocument.documentElement.clientWidth;
```
simshanith:
```suggestion
      // Measure against the document's client width to account for scrollbar.
      const layoutWidth = $content[0].ownerDocument.documentElement.clientWidth;
```

packages/lit-ui-router-mobx/vitest.config.ts
```diff
@@ -3,6 +3,10 @@ import { defineConfig } from 'vitest/config';
 import { playwright } from '@vitest/browser-playwright';
 
 export default defineConfig({
+  // test and test:coverage run concurrently in this directory; sharing
+  // node_modules/.vite lets their dep-optimizer runs corrupt each other
+  // (browsers collect 0 tests, or the run never finishes).
+  cacheDir: `node_modules/.vite-${process.env.VITEST_BROWSER_API_PORT ?? 'default'}`,
```
simshanith:
```suggestion
  cacheDir: `node_modules/.vite-${process.env.VITEST_BROWSER_API_PORT ?? 'default'}`,
```

packages/lit-ui-router-mobx/vitest.config.ts
```diff
@@ -15,6 +19,10 @@ export default defineConfig({
   test: {
     globals: true,
     include: ['src/specs/**/*.spec.ts'],
+    // vitest 4.1.x browser mode intermittently finishes green but never
+    // exits on ubuntu runners; hanging-process dumps the open handles so
+    // the CI log shows what kept the process alive.
+    reporters: process.env.CI ? ['default', 'hanging-process'] : ['default'],
```
simshanith:
```suggestion
    // hanging-process logs the open handles in CI
    reporters: process.env.CI ? ['default', 'hanging-process'] : ['default'],
```

packages/lit-ui-router-mobx/vitest.config.ts
```diff
@@ -23,6 +31,13 @@ export default defineConfig({
     browser: {
       enabled: true,
       headless: true,
+      // Every concurrent vitest browser process defaults to port 63315 and
+      // races the fallback rebind; a loser's browser can attach to the wrong
+      // server (0 tests collected, or the run waits forever — the CI hang).
+      // Each test script pins a distinct port via VITEST_BROWSER_API_PORT.
+      api: process.env.VITEST_BROWSER_API_PORT
+        ? { port: Number(process.env.VITEST_BROWSER_API_PORT) }
+        : undefined,
```
simshanith:
```suggestion
      api: process.env.VITEST_BROWSER_API_PORT
        ? { port: Number(process.env.VITEST_BROWSER_API_PORT) }
        : undefined,
```

Issue #11, opened 2025-07-24
Benchmark/ perf testing


nickschaap:
We should consider doing some amount of perf testing to protect against regressions and also be able to report performance. Not sure what that looks like exactly. Creating this issue to track.

PR #183, opened 2026-07-07
ci(codecov): bundle analysis for sample app builds

simshanith:
Part of #11

The "easy win" slice of perf/regression testing: Codecov bundle analysis, wired via [`@codecov/vite-plugin`](https://docs.codecov.com/docs/javascript-bundle-analysis) (v2.0.1, added to the workspace catalog).

## What's tracked

The publishable packages build with plain `tsc` (unbundled ESM), so there is no library bundle to instrument directly. The two vite-built sample apps are the real bundles, and together they cover all three libraries:

| Bundle | Libraries exercised |
| --- | --- |
| `sample-app-lit-vanilla` | `lit-ui-router`, `ui-router-navigation-location-plugin` (via `sample-app-shared`) |
| `sample-app-lit-mobx` | the above + `lit-ui-router-mobx` |

Library `src/` changes propagate through turbo (`dependsOn: ^build`) into fresh app builds, so library size regressions show up as bundle deltas on every PR.

## How it behaves

- **Local / no token:** `enableBundleAnalysis` is gated on `CODECOV_TOKEN`, so the plugin is a complete no-op — builds don't fail and nothing is uploaded. Verified with a tokenless `turbo run build`.
- **CI:** the existing `CODECOV_TOKEN` secret (already used by `codecov/codecov-action`) is passed to the `turbo run ci` step; the plugin uploads bundle stats during `vite build`.
- **Turbo env plumbing:** `CODECOV_TOKEN` is declared in the apps' `build.env` (hash-affecting, so tokenless local cache artifacts can't satisfy CI builds); `GITHUB_*` is in `passThroughEnv` (available for the plugin's CI/commit detection under strict env mode, without busting the cache every commit). Note: on a full turbo cache hit the build (and upload) is skipped — acceptable, since a cache hit means the bundle inputs didn't change.
- `telemetry: false` keeps the plugin from sending Sentry telemetry; `gitService: 'github'` matches the documented public-repo GitHub Actions setup.

## Follow-up (dashboard side)

- Bundle analysis appears under the repo's **Bundles** tab on Codecov after the first upload from `main`; no extra Codecov configuration is required, but the tab stays empty until a tokened build runs.
- Optionally add `bundle_analysis` status/comment settings to `codecov.yml` later if PR comments on bundle deltas are wanted.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -58,6 +58,8 @@ jobs:
           TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
           TURBO_API: ${{ vars.TURBO_API }}
           TURBO_TEAM: ${{ vars.TURBO_TEAM }}
+          # Codecov bundle analysis (uploaded by @codecov/vite-plugin during app builds)
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
       - name: lit-ui-router Vitest Coverage Report
         uses: davelosert/vitest-coverage-report-action@v2
         if: github.event_name == 'pull_request'
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -58,7 +58,6 @@ jobs:
           TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
           TURBO_API: ${{ vars.TURBO_API }}
           TURBO_TEAM: ${{ vars.TURBO_TEAM }}
-          # Codecov bundle analysis (uploaded by @codecov/vite-plugin during app builds)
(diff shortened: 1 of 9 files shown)


.github/workflows/build-test.yml
```diff
@@ -58,6 +58,8 @@ jobs:
           TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
           TURBO_API: ${{ vars.TURBO_API }}
           TURBO_TEAM: ${{ vars.TURBO_TEAM }}
+          # Codecov bundle analysis (uploaded by @codecov/vite-plugin during app builds)
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
```suggestion
          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```

PR #267, opened 2026-07-10
test(examples): typecheck the standalone examples against the published package

simshanith:
## Why

A tsconfig-coverage audit of the repo (simulating #260 + #263 merged) found exactly six `.ts` files not typechecked by any CI-exercised tsconfig — all under `examples/`:

- `examples/{helloworld,hellosolarsystem,hellogalaxy}/src/main.ts` — bundled by `build:embeds` and **shipped to lit-ui-router.dev as embeds**, but only ever esbuild-transpiled (types erased, errors silently tolerated)
- `examples/{helloworld,hellosolarsystem,hellogalaxy}/vite.config.ts` — esbuild-loaded by vite, and not even in the examples' tsconfig `include`

Each example has a `tsconfig.json`, but it was **false coverage**: the examples are standalone npm projects, not pnpm workspace members (`pnpm-workspace.yaml` lists `examples`, not `examples/*`), so no turbo task ever ran `tsc` against them — their `build` is a bare `vite build`.

## What

- `examples/package.json` gets a `typecheck` script running the three `tsc -p <example>` invocations via `concurrently` (matching the `test:cypress:all` pattern), with `typescript: catalog:` and `concurrently: catalog:` devDeps.
- `concurrently` moves to the pnpm catalog (`check:catalog` requires it once two packages declare it inline); `apps/sample-app-lit-e2e` switches to the catalog ref.
- `examples/turbo.json` adds `typecheck` with `dependsOn: ["build:embeds"]` — the examples resolve the published `lit-ui-router` from their own `npm ci` node_modules, which `build:embeds` installs. Inputs are inherited (`$TURBO_DEFAULT$`), nothing enumerated.
- Each example tsconfig widens `include` to `["src/**/*", "vite.config.ts"]` and gains an explicit `"types": []` per repo convention.

## Semantics: consumer-floor checking, standalone design preserved

The examples deliberately keep their standalone-npm-project design: they typecheck against the **published** `lit-ui-router` from the registry (via their own `package-lock.json`), not the workspace build. That means a type-breaking change in `packages/lit-ui-router` surfaces here only after release — the same philosophy as `tools/dts-backtest`. No `paths` mapping, no workspace coupling.

## Verification

- `turbo run typecheck --filter=examples` — all three examples pass under TS 7.0.2; `examples#typecheck` executes after `examples#build:embeds`.
- Falsifiability: a planted type error in `hellosolarsystem/src/main.ts` fails the task; a planted error in `helloworld/vite.config.ts` also fails it (proving the include widening is load-bearing); reverted, clean run passes.
- Cache: unchanged rerun is a cache hit; editing `vite.config.ts` is a cache miss without `--force` (inherited `$TURBO_DEFAULT$` inputs over-approximate).
- `turbo run ci --dry=json` confirms `examples#typecheck` is in the `ci` graph with `dependencies: ["examples#build:embeds"]`.
- `turbo run ci --filter=examples` — 10/10 tasks green, including `//#check:catalog`, `//#check:pack`, `//#lint:package-json`, `//#test:root`, and format checks.

## Hermeticity (bd8d132, reworked in 667d50e)

`examples#typecheck` originally leaned on `build:embeds` having *executed* `npm ci` — but a turbo **cache hit** on `build:embeds` restores only its declared `*/dist/**` outputs, not `node_modules/`, so typecheck failed with `TS2307: Cannot find module 'lit'/…` whenever the cache was warm (as on head d48f429 in CI). Per review, install logic now lives entirely in the examples package's `postinstall` hook: three self-contained `example:install:<example>` scripts (`npm ci --no-audit --no-fund --prefix <example>`) run concurrently via `concurrently "npm:example:install:*"`, legs auto-named from the wildcard suffix — no install code in runtime TS. `build-embeds.ts` is a pure vite-build step and `typecheck:<example>` is pure `tsc`; the `typecheck -> build:embeds` turbo edge (which only served to serialize installs) is dropped.

- pnpm 11 fires workspace projects' own lifecycle scripts without extra config (its script gating applies to dependencies) — verified empirically on a fresh `pnpm install --frozen-lockfile`, which is exactly what `build_and_test` runs. #269's commitlint job uses `--ignore-scripts`, which is fine (it doesn't need the examples).
- Caveat: pnpm's up-to-date fast path skips lifecycle, so a no-op `pnpm install` won't restore manually deleted `examples/*/node_modules`; any real install does, or run `pnpm --filter examples postinstall` directly (documented in `examples/README.md`). No-op installs stay at ~0.3s; a real install adds ~3s of concurrent `npm ci` with a warm npm cache.
- Caching `*/node_modules/**` as a build:embeds output was rejected: platform-specific esbuild/vite binaries don't survive cross-runner cache restores.
- Verified: wipe examples node_modules -> `pnpm install` fires the hook and restores all three; `turbo run typecheck --filter=examples` passes standalone with zero install-at-task-time; `turbo run build:embeds` builds all three with no `npm ci` in its output; `turbo run ci --filter=examples` 10/10 green.


examples/turbo.json
```diff
@@ -13,6 +13,11 @@
         "*/vite.config.ts"
       ],
       "outputs": ["*/dist/**"]
+    },
+    "typecheck": {
+      // The examples typecheck against the published lit-ui-router in their
+      // own npm-ci node_modules, which build:embeds installs.
```
simshanith:
```suggestion
```

simshanith: Applied in 5a80a68 — comment dropped, bare graph edge kept.

examples/package.json
```diff
@@ -7,6 +7,11 @@
     "build:embeds": "node build-embeds.ts",
     "format": "oxfmt \"**/*.{ts,js,json,md,html}\"",
     "format:check": "oxfmt --check \"**/*.{ts,js,json,md,html}\"",
-    "lint": "oxlint ."
+    "lint": "oxlint .",
+    "typecheck": "concurrently --names helloworld,hellosolarsystem,hellogalaxy --prefix-colors auto \"tsc -p helloworld\" \"tsc -p hellosolarsystem\" \"tsc -p hellogalaxy\""
```
simshanith: split into three `typecheck:hello*` and concurrently call the package scripts, like other examples

simshanith: Done in d48f429 — three `typecheck:hello*` scripts, with `typecheck` concurrently calling them via `pnpm run` (same shape as `test:cypress:all`).

examples/package.json
```diff
@@ -7,6 +7,14 @@
     "build:embeds": "node build-embeds.ts",
     "format": "oxfmt \"**/*.{ts,js,json,md,html}\"",
     "format:check": "oxfmt --check \"**/*.{ts,js,json,md,html}\"",
-    "lint": "oxlint ."
+    "lint": "oxlint .",
+    "typecheck": "concurrently --names helloworld,hellosolarsystem,hellogalaxy --prefix-colors auto \"pnpm run typecheck:helloworld\" \"pnpm run typecheck:hellosolarsystem\" \"pnpm run typecheck:hellogalaxy\"",
+    "typecheck:hellogalaxy": "node build-embeds.ts --install hellogalaxy && tsc -p hellogalaxy",
+    "typecheck:hellosolarsystem": "node build-embeds.ts --install hellosolarsystem && tsc -p hellosolarsystem",
+    "typecheck:helloworld": "node build-embeds.ts --install helloworld && tsc -p helloworld"
```
simshanith: ci and local should install, perhaps as a postinstall hook so pnpm install -> automatically sets up examples concurrently?

simshanith: Done in bd8d132 — `postinstall` on the examples package runs the three `build-embeds.ts --install <example>` steps concurrently, so `pnpm install` sets them up; `typecheck:hello*` is pure `tsc` again. Note pnpm's up-to-date fast path skips lifecycle scripts, so a no-op install won't restore manually deleted `examples/*/node_modules` — any real install (fresh clone, CI, lockfile change) fires it.

simshanith: Addendum: the escape hatch is now documented in `examples/README.md` (9f36993) — `pnpm --filter examples postinstall` restores wiped example node_modules; skipped an alias since it wouldn't be shorter than the raw command and `install` is a reserved lifecycle name.

examples/build-embeds.ts
```diff
@@ -11,10 +11,19 @@ import { fileURLToPath } from 'node:url';
...
-for (const name of EXAMPLES) {
+// `--install <name>` runs only the npm ci step, for tasks (typecheck) that
+// need the example's node_modules but can't rely on a cached build:embeds.
+const installOnly = process.argv[2] === '--install';
+const only = installOnly ? process.argv[3] : undefined;
+if (installOnly && (only === undefined || !EXAMPLES.includes(only))) {
+  throw new Error(`--install expects one of: ${EXAMPLES.join(', ')}`);
+}
+
+for (const name of only ? [only] : EXAMPLES) {
```
simshanith: prefer moving out of runtime ts to postinstall package.json directly

```sh
npm ci --no-audit --no-fund
```

simshanith: with `--prefix` per directory

simshanith: Done in 667d50e — `postinstall` now runs `npm ci --no-audit --no-fund --prefix <example>` per directory via concurrently, and build-embeds.ts is a pure vite-build step (no install logic in runtime TS). The now-purposeless `typecheck -> build:embeds` turbo edge is dropped too.

simshanith: Addendum: 69babc1 dedupes the flags behind `"example:install": "npm ci --no-audit --no-fund --prefix"` — postinstall's legs are `pnpm run example:install <example>` (pnpm appends the arg, verified the exact npm invocation). Also handy solo: `pnpm --filter examples example:install <example>`.

simshanith: Final shape per owner in 33c98bb: three self-contained `example:install:<example>` scripts and `postinstall: concurrently \"npm:example:install:*\"` — each runnable alone (`pnpm --filter examples example:install:hellogalaxy`), legs auto-named from the wildcard suffix. Verified through the pnpm lifecycle: wipe + `pnpm install` restores all three.

PR #268, opened 2026-07-10
feat(tooling): dashboard-as-code check for the Workers Builds triggers

simshanith:
## Motivation

The Workers Builds configuration that deploys [lit-ui-router.dev](https://lit-ui-router.dev) lives only in the Cloudflare dashboard — no PR trail, no drift detection. This PR adds a dashboard-as-code check plus an empirical investigation of what the deploy/version commands could become without the docs `wrangler` package.json alias (context: #266).

## Part A — `tools/workers-builds` (private package)

Reviewable source of truth (`DESIRED` constant) for the two Workers Builds triggers, diffed against the live API.

```bash
CLOUDFLARE_API_TOKEN=… CLOUDFLARE_ACCOUNT_ID=… pnpm check:workers-builds             # read-only diff
CLOUDFLARE_API_TOKEN=… CLOUDFLARE_ACCOUNT_ID=… pnpm check:workers-builds -- --apply  # PATCH drifted triggers, re-GET to confirm
```

- Env: `CLOUDFLARE_API_TOKEN` (user token, **Workers Builds Configuration: Edit**) and `CLOUDFLARE_ACCOUNT_ID`, both required — fails fast naming both; the token is never logged.
- Worker name comes from the root `wrangler.jsonc` `name` field (minimal string-aware JSONC parse, no new deps).
- Flow: `GET /workers/scripts` → worker tag → `GET /builds/workers/{tag}/triggers` → classify production (builds `main`) vs preview → diff `build_command` / `deploy_command`; `root_directory` and branch config are printed but not pinned (the dashboard's root-directory setting is unknown, and command forms were verified to work under either — see Part B).
- Exit codes: `0` in sync, `1` drifted (or drift remains after `--apply`), `2` usage/API error. Read-only unless `--apply`.
- **The desired state lives in `tools/workers-builds/workers-builds-triggers.config.jsonc` and mirrors the CURRENT dashboard values** (build `npx turbo docs#build`, deploy `pnpm wrangler deploy`, preview `pnpm wrangler versions upload`); the recommended alias-free alternatives sit next to each field as JSONC comments. JSONC because `jsonc-parser` is already the script's parser and comments preserve those alternatives; the file is loaded and strictly validated at runtime (missing/invalid/typoed keys fail fast, exit 2) with the validation in the tested core.
- Follows the #260 `.ts` script convention (node 24 type stripping, IO shell + pure `*.core.ts` + `node --test`); the `@types/node` catalog entry and oxlint `scripts/**/*.test.ts` override are identical to that branch, so the eventual merge is a no-op.
- Deliberately **not** wired into any CI/turbo task graph — it reads (and with `--apply`, writes) production deploy configuration.

Tested without credentials: 12 new `node --test` cases (JSONC parsing against the real `wrangler.jsonc`, trigger classification, drift/patch-body construction, missing-trigger reporting), plus fail-fast verification with envs unset (exit 2, clear message).

## Part B — alias-free command matrix (wrangler 4.107.0, empirical)

Key finding: wrangler discovers `wrangler.jsonc` by **walking up from the cwd** and resolves `assets.directory` relative to the config file — so every form below read the root config and picked up `docs/dist` from both candidate root directories.

| Command form | cwd = repo root | cwd = `docs/` | Notes |
| --- | --- | --- | --- |
| `pnpm wrangler --version` | works | works (alias) | |
| `pnpm wrangler deploy --dry-run` | works (exec fallback) | works — **verified with the alias removed** (falls back to `pnpm exec`) | current dashboard value |
| `pnpm wrangler versions upload --dry-run` | works | (same mechanics) | current dashboard value |
| `pnpm exec wrangler deploy --dry-run` | works | works | upward config discovery |
| `pnpm --filter docs exec wrangler deploy --dry-run` | works | works | pins cwd to `docs/` |
| `pnpm --filter docs exec wrangler versions upload --dry-run` | works | works | |
| `npx --no-install wrangler deploy --dry-run` | works | works | resolves local binary, no download |

**Recommendation per dashboard field** (robust to either root-directory setting):

1. Keep `pnpm wrangler deploy` / `pnpm wrangler versions upload` — already alias-free-safe: with the docs alias deleted, pnpm's exec fallback resolves the workspace-pinned binary and config discovery does the rest (verified empirically).
2. `pnpm --filter docs exec wrangler deploy` / `… versions upload` — equally correct, and additionally pins the cwd to `docs/`, immune to any future root-directory change.
3. `npx wrangler …` works today but is disqualified as a recommendation: if the `wrangler` devDependency were ever removed, `npx` would silently fetch an unpinned latest wrangler.

DEPLOY.md gains this matrix (condensed) next to the existing commands table, plus a "Dashboard as Code" section documenting the script.

## Notes

- **Layout (user-requested pre-merge feedback)**: the checker, its tests, and the desired-state config were restructured from loose `scripts/` files into the private `tools/workers-builds` package, following the existing `tools/*` conventions (own `lint`/`format:check`/`test`/`typecheck` tasks under `turbo run ci` — package-level lint matters because `lint:root` ignores `tools/`). Root `check:workers-builds` runs `turbo run workers-builds#check` (matching the other turbo-backed root scripts); the package's own `turbo.json` (`extends: ["//"]`) owns the `check` task (`cache: false`, `passThroughEnv` for the credentials) and the `test` inputs override. Turbo propagates the task exit code unflattened, so the 0/1/2 contract survives; `--apply` is passed as `pnpm check:workers-builds -- --apply`.

- `wrangler.jsonc` is parsed with [`jsonc-parser`](https://www.npmjs.com/package/jsonc-parser) (Microsoft's VS Code JSONC implementation, new root devDep via the catalog); parse errors surface with offset + `ParseErrorCode`.
- Test expectations (API trigger shapes, drift scenario) live in `tools/workers-builds/workers-builds-triggers.fixtures.json`, doubling as documentation of the Builds API shapes; the package's `test` turbo task uses `$TURBO_DEFAULT$` inputs so the fixture and config are tracked.
- Retirement path: this script should be replaced by the official Cloudflare SDK / Terraform / Pulumi resource once the Workers Builds API lands in Cloudflare's OpenAPI spec — verified today (2026-07-09) that none of them cover it yet.
- DEPLOY.md carries only a terse Dashboard-as-Code note; the alias-free matrix above is preserved here in the PR body as review evidence rather than living documentation.

- The config intentionally does not adopt the alias-free recommendation — the dashboard has not been changed; flipping it is a follow-up decision.
- `test:root` now globs both `*.test.mjs` and `*.test.ts`; #260 rewrites that line anyway.


scripts/workers-builds-triggers.ts
```diff
@@ -0,0 +1,165 @@
...
+  },
+  preview: {
+    build_command: 'npx turbo docs#build',
+    deploy_command: 'pnpm wrangler versions upload',
+  },
+  // Alias-free alternatives (verified in DEPLOY.md — work from either root
+  // directory, with or without the docs `wrangler` package.json script):
+  //   production: { deploy_command: 'pnpm --filter docs exec wrangler deploy' }
+  //   preview: { deploy_command: 'pnpm --filter docs exec wrangler versions upload' }
+};
```
simshanith: source of truth should live outside the script in own config.json(c) or yaml or toml or wherever, not hardcoded in script

simshanith: Done in bf593943e2e12ac28b42f664a97826db67fde9cc — the desired state now lives in `scripts/workers-builds-triggers.config.jsonc` (JSONC so the commented alias-free alternatives carry over, parsed by the `jsonc-parser` dependency the script already uses). Loaded and strictly validated at runtime; unknown/typoed keys fail fast with exit 2, and the tests exercise the real config file on every `test:root` run.

scripts/workers-builds-triggers.config.jsonc
```diff
@@ -0,0 +1,20 @@
+// Desired Workers Builds trigger configuration for lit-ui-router.dev — the
+// source of truth the Cloudflare dashboard should match. Mirrors the CURRENT
+// dashboard values; change this file first, then run
+// `pnpm check:workers-builds --apply`.
```
simshanith:
```suggestion
// Desired Workers Builds trigger configuration
// `pnpm check:workers-builds --apply` updates Cloudflare
```

scripts/workers-builds-triggers.config.jsonc
```diff
@@ -0,0 +1,20 @@
...
+{
+  "productionBranch": "main",
+  "production": {
+    "build_command": "npx turbo docs#build",
+    "deploy_command": "pnpm wrangler deploy",
+    // Alias-free alternative (verified in PR #268 — works from either root
+    // directory, with or without the docs `wrangler` package.json script):
+    // "deploy_command": "pnpm --filter docs exec wrangler deploy",
```
simshanith:
```suggestion
```

scripts/workers-builds-triggers.config.jsonc
```diff
@@ -0,0 +1,20 @@
...
+  },
+  "preview": {
+    "build_command": "npx turbo docs#build",
+    "deploy_command": "pnpm wrangler versions upload",
+    // Alias-free alternative:
+    // "deploy_command": "pnpm --filter docs exec wrangler versions upload",
```
simshanith:
```suggestion
```

simshanith (commented):
alias free alternatives determined unnecessary iirc; simplify comments

scripts/workers-builds-triggers.config.jsonc
simshanith: Applied in a3dcda2fdab62d5bb86b218eec40d8010dc61566.

scripts/workers-builds-triggers.config.jsonc
simshanith: Applied in a3dcda2fdab62d5bb86b218eec40d8010dc61566.

scripts/workers-builds-triggers.config.jsonc
simshanith: Applied in a3dcda2fdab62d5bb86b218eec40d8010dc61566.

Issue #13, opened 2025-07-24
Add commitlint / changeset


nickschaap:
This isn't as important now but especially when we start getting ready to release we should have commitlint/ changeset enabled

simshanith:
https://www.conventionalcommits.org/en/v1.0.0/
https://github.com/conventional-changelog/commitlint
https://github.com/release-it/conventional-changelog

PR #269, opened 2026-07-10
ci(github): lint PR commits with commitlint

simshanith:
Part of #13. Follows #185 (title lint, merged) and #207 (changelog, merged).

## What it does

Adds a `lint_pr_commits` check: `wagoid/commitlint-github-action@v6` linting every commit on the PR branch against `commitlint.config.mjs` (`@commitlint/config-conventional`). With squash merges configured as `PR_TITLE`/`COMMIT_MESSAGES`, these messages become the squash commit body on `main` — this check is the only format gate on what lands in git history, including the `BREAKING CHANGE:` footers that signal majors.

## Why the standard action (not a custom script)

An earlier revision built a bespoke Octokit + programmatic-commitlint script. Swapped for the ecosystem-standard action to drop the maintenance burden: no root deps beyond the CLI kept for local use, no script/test/tsconfig surface to own, and the action tracks commitlint upstream. The tradeoff that originally motivated the script — base-trusted config under `pull_request_target` — is deliberately accepted instead: the job runs on the **unprivileged `pull_request` event** (no secrets, read-only token), where checking out PR code is safe by construction. A PR could edit its own config/workflow to pass, but that's inherent to any `pull_request`-based lint and gates nothing it could steal.

## Workflow shape: two files, one trigger each

The title lint stays in `semantic-pr.yml` on `pull_request_target`, untouched by this PR. The commit lint gets its own `commitlint.yml` on `pull_request`, with the same `release`-label exemption (bump-version PRs carry `Release X.Y.Z` commits by design) and its own per-PR concurrency group. Why not one file with both triggers: a shared file runs every job under both events, and the skipped twins still report check runs under the same name — GitHub treats a skipped conclusion as *satisfying* a required status check, so once these checks join the `main` ruleset, a no-op twin could green-light a merge the real check never validated. With one trigger per file, the `lint_pr_commits` name only ever comes from the event that actually runs the lint, and a missing run blocks as "Expected". The check name is stable for a future ruleset requirement.

## Exemptions (in `commitlint.config.mjs`, shared with local runs)

commitlint's default ignores already cover GitHub's `Merge branch/pull request/remote-tracking branch` wordings, `Revert "…"`, and `fixup!`/`squash!` (verified by echoing each through `pnpm exec commitlint`). One custom ignore added for hand-typed lowercase `merge <x> into <y>` freshens, which the defaults miss. `Release 1.6.1` fails commitlint, so the label exemption is load-bearing. Locally: `echo "feat: …" | pnpm exec commitlint`.

## Self-validating

Because `pull_request` runs the PR branch's copy of `commitlint.yml`, this check runs on this very PR — its own commits are the first real input.

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -1,0 +1,41 @@
+# Lints the PR's branch commits: with squash merges configured as
+# PR_TITLE/COMMIT_MESSAGES, they become the squash commit body on main.
+# Kept separate from semantic-pr.yml so the lint_pr_commits check name is
+# only ever produced by this trigger - a skipped twin job under another
+# event would still satisfy a required status check.
+name: Commitlint
+
+on:
+  # pull_request runs the PR-branch workflow file with no secrets, so
+  # checking out and reading the PR's own commits is safe.
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      # labeled/unlabeled: the release-label exemption must re-run the check
+      - labeled
+      - unlabeled
+
+# Re-validate only the latest commits per PR
+concurrency:
+  group: commitlint-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  lint_pr_commits:
+    # Release PRs carry "Release X.Y.Z" commits by design - same exemption
+    # as the title lint's ignoreLabels.
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'release') }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          # Full history so the action can walk the PR's commit range
+          fetch-depth: 0
+      # Lints with the repo's commitlint.config.mjs (merge/revert/fixup
+      # exemptions live there; the action loads only ESM .mjs configs)
+      - uses: wagoid/commitlint-github-action@v6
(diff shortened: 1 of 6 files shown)


.github/workflows/semantic-pr.yml
```diff
@@ -72,3 +74,46 @@ jobs:
...
+          status=0
+          for i in $(seq 0 "$((count - 1))"); do
+            sha=$(jq -r ".[$i].sha" /tmp/commits.json)
+            jq -r ".[$i].commit.message" /tmp/commits.json > /tmp/message.txt
+            if ! pnpm exec commitlint < /tmp/message.txt; then
+              echo "::error::Commit $sha is not a conventional commit (see commitlint output above)."
+              status=1
+            fi
+          done
+          exit "$status"
```
simshanith: extract to scripts/pr-commitlint.ts; consider using octokit

simshanith: Done in 27c36bf — extracted to `scripts/pr-commitlint.ts` (IO shell: Octokit fetch + config load) and `scripts/pr-commitlint.core.ts` (pure, unit-tested lint/report logic), following the check-catalog core/shell convention. Swapped `gh api | jq` for `@octokit/rest` paginate and the commitlint CLI spawn for programmatic `@commitlint/load` + `@commitlint/lint`. The workflow step is now a single `node scripts/pr-commitlint.ts`; messages still never pass through shell interpolation (script receives only repo + PR number via env). Re-ran the real-PR matrix through the new path — #207/#266/#267 all pass with identical verdicts.

PR #227, opened 2026-07-07
chore(ci): reduce build-and-test log noise

simshanith:
Follow-up to the CI-log audit of [run 28891367836](https://github.com/simshanith/lit-ui-router/actions/runs/28891367836/job/85704648154). Companion to #226 (allowCypressEnv, merged). Changes:

1. **wrangler request logging (666 lines)** — `WRANGLER_LOG: warn` on the Build and Test step silences per-request `[wrangler:info]` lines from the e2e dev server.
2. **"Lit is in dev mode" banner (36×, once per Vitest worker)** — seed `globalThis.litIssuedWarnings` with `'dev-mode'` in a new `vitest.setup.ts` for `packages/lit-ui-router`. This is lit's sanctioned per-warning switch (lit/lit#4877); other warnings still surface.
3. **Duplicate "first time using Cypress" verification banner** — run `cypress verify` in the Install Cypress step so the concurrent vanilla/mobx runs find existing verification state and skip it.
4. **Cypress binary download during `pnpm install`** — `CYPRESS_INSTALL_BINARY: '0'` on the dependency install step; the dedicated Install Cypress step owns the download.
5. **Xvfb xkbcomp keysym warnings (18 lines)** — startup extracted to `scripts/start-xvfb.mjs` (scripts/ convention: `.core.mjs` logic + `node:test` coverage, unit and e2e against fake Xvfb binaries). Stderr goes to an unlinked temp file — a durable sink Xvfb can keep writing to after the step ends, where a pipe would EPIPE on later client connects. The script polls for the display socket; on startup failure it prints the capture with only the exact `> Warning: Could not resolve keysym` lines filtered — noting how many warnings were filtered — and exits nonzero. Real errors (fatal server errors, other xkbcomp output) always surface.
6. **Vite 500 kB chunk warning in `examples:build:embeds`** — bump `chunkSizeWarningLimit` to 1100 in `examples/hellogalaxy/vite.config.ts`; the ~1 MB `model-viewer` chunk is fine for a demo embed.
7. **Codecov gcov/coverage.py probe warnings** — `disable_search: true`, with the `files:` list expanded to the six reports auto-search actually uploaded (lcov + istanbul json × 3 packages). The earlier two-file list dropped mobx coverage entirely (project 97.71% → 93.08%) and lcov-alone reported more partial branches.
[...]

PR #273, opened 2026-07-10
ci(lint): repo-owned actionlint workflow

simshanith:
Delivers the actionlint follow-up promised in #232, upgraded per discussion from local-only to a standalone CI workflow plus a `ci`-task second net.

## Why now

#227, #253, and #259 each claimed "actionlint is clean" using a **global** binary the repo doesn't own — exactly the environment-divergence trap #232 documented for prettier. This PR makes the actionlint version the repo's business: `.config/mise/config.toml` is the source of truth, and `mise which actionlint` resolves to the project-pinned tool by construction.

## Provisioning: mise aqua backend, pinned in `.config/mise/config.toml`

```toml
"aqua:rhysd/actionlint" = "1.7.12"
```

rhysd publishes no official npm distribution, so earlier revisions of this branch tried the `github-actionlint` npm wrapper, then the mise **go** backend. Per owner preference the tool is a project mise tool; between the two mise backends, **aqua wins on the evidence**:

| axis | `aqua:rhysd/actionlint` | `go:github.com/rhysd/actionlint/cmd/actionlint` |
| --- | --- | --- |
| install | official prebuilt release binary; ~0.4 s reinstall locally | ~16 s `go install` build in CI, plus Go toolchain switching for the module's `go >= 1.25` floor |
| mise.lock | sha256 checksums for **all 7 platforms** written on add (linux runners covered; reinstall from the committed lock is byte-identical — no platform stripping) | version only; integrity delegated to the Go module proxy sumdb at build time |
| supply chain | mise verifies **GitHub artifact attestations** at install (`provenance = "github-attestations"` in the lock) | source build via proxy + sumdb; nothing pinned in-repo |
| pinning | exact `1.7.12` | exact `1.7.12` (spec must target `cmd/actionlint`; the module root is a library) |
| precedent | pnpm is already aqua-backed in this repo | — |

Provenance proof: the mise tool's `--version` banner reads `installed by downloading from release page` at `~/.local/share/mise/installs/aqua-rhysd-actionlint/1.7.12/`, and the shim outranks `/opt/homebrew/bin/actionlint` on `PATH`.

## Workflow design

Final standalone workflow is three steps — **checkout → mise-action → `actionlint`** — no `pnpm install`, no node_modules, no secrets (jdx/mise-action installs the pinned, checksum-verified binary and puts it on `PATH`).

- **Standalone on purpose**: a PR editing `build-test.yml` is linted by a workflow that edit cannot break; a syntactically invalid `build-test.yml` (which GitHub refuses to run) still gets a failing lint check instead of silence.
- **Also wired into `turbo run ci`** via the root `lint` task's `with` array (the repo's existing root-task pattern): double-checking from both workflows is intended — build_and_test (which already runs mise-action) re-runs the lint, and local `turbo run ci` catches workflow edits before push (`--dry=json` shows exactly one `//#lint:workflows` occurrence). The task is cacheable with over-approximating inputs (`.github/**` + root `$TURBO_DEFAULT$`); verified a workflow-file edit flips the task hash to a cache miss and reverting restores the hit.
- The root script is an explicit `mise exec -- actionlint`, so local runs resolve the project-pinned tool regardless of shim PATH order.
- **Triggers**: `pull_request` + `push` to `main`. **No `paths` filter**: the run is cheap (mise-action cached, lint <1 s), and a path-filtered workflow reports no status at all on unrelated PRs — which would deadlock merges if it were ever made a required check.

## shellcheck

actionlint lints `run:` blocks with shellcheck only when a binary is on `PATH`; mise provisions actionlint but **not** shellcheck. Local verification ran with shellcheck 0.11.0 present, and a scratch workflow with an unquoted expansion produced SC2086 through the project-pinned binary, confirming the integration fires. ubuntu-latest runner images ship shellcheck, so both CI nets get the same coverage; a contributor without shellcheck gets a quieter (but still valid) actionlint run.

## Findings on existing workflows

**None.** actionlint 1.7.12 (with shellcheck) exits 0 over `build-test.yml`, `bump-version.yml`, `publish-gh.yml`, `publish-npm.yml`, and the new `lint-workflows.yml`.

## Verification

- aqua lock delta is purely additive; all node/pnpm entries untouched; `mise uninstall` + `mise install` against the committed lock verifies the pinned checksum and leaves the lock byte-identical
- `turbo run ci --dry=json` lists `//#lint:workflows` (115 tasks)
- local gate: `turbo run lint typecheck format:check '//#check:catalog' '//#check:pack' '//#test:root' '//#lint:workflows'` → 52/52 successful


.github/workflows/lint-workflows.yml
```diff
@@ -0,0 +1,30 @@
+# Lints .github/workflows with the project-pinned actionlint (mise aqua backend).
+# Standalone on purpose: kept out of build_and_test so edits to that workflow
+# are themselves linted by a workflow the edit cannot break. build_and_test
+# runs the same lint again via `turbo run ci` as a second net.
```
simshanith:
```suggestion
# Standalone actionlint workflow
```

package.json
```diff
@@ -17,6 +17,7 @@
     "lint": "turbo run lint",
     "lint:package-json": "eslint --format tap \"**/package.json\"",
     "lint:root": "oxlint --ignore-pattern apps --ignore-pattern tools --ignore-pattern docs --ignore-pattern packages --ignore-pattern examples .",
+    "lint:workflows": "actionlint",
```
simshanith:
```suggestion
    "lint:workflows": "mise exec -- actionlint",
```

consider explicit `mise exec`

turbo.json
```diff
@@ -60,7 +63,8 @@
         "format:check",
         "codecov:bundle",
         "//#check:catalog",
-        "//#check:pack"
+        "//#check:pack",
+        "//#lint:workflows"
```
simshanith: prefer adding to `lint: { with: [] }`

.github/workflows/lint-workflows.yml
simshanith: Applied verbatim in 416939c.

package.json
simshanith: Applied in 416939c — `mise exec -- actionlint`; verified `pnpm run lint:workflows` and the turbo task still resolve the project-pinned binary.

turbo.json
simshanith: Rewired in 416939c: `//#lint:workflows` now rides `lint`'s `with` array and the `ci` dependsOn edge is gone. `turbo run ci --dry=json` shows exactly one occurrence, and a workflow-file edit still busts the cache (hit → miss → hit on revert).

PR #274, opened 2026-07-10
fix(e2e): make hash mode green and run it in CI

simshanith:
## Summary

Hash routing is a published configuration of `ui-router-navigation-location-plugin`, but CI never ran it (#226: "Nothing in CI sets this today"), and #252 documented that `feature_flags_panel.cy.js` fails identically on `main` under `LOCATION_PLUGIN=hash`. This PR fixes the two failing specs and folds hash- and navigation-mode e2e legs into the `test` task (so `turbo run ci` and `build_and_test` cover them) so the configurations can't rot again.

## Root cause: test harness, not product

Reproduced on unmodified `main` (full vanilla suite, `--expose LOCATION_PLUGIN=hash`): **25/27**, both `feature_flags_panel.cy.js` tests failing with

```
CypressError: Timed out retrying after 4000ms: `cy.select()` failed because this element is `disabled`:
`<select disabled="">...</select>`
```

The chain:

1. `visitWithFeatures` injected the suite-wide `LOCATION_PLUGIN` as `?feature-location-plugin=hash` in `location.search`.
2. The hash plugin only ever rewrites `location.hash`, so the param persists for the entire session (under pushState it's dropped on the first router navigation).
3. `featureFlags.isUrlOverridden('location-plugin')` therefore stays `true`, and `FeatureFlagsPanel` renders the select `disabled` — which is *correct* app behavior: while the URL param is present, a panel selection really would be ignored.

So the library and the sample app behave as designed; the harness was configuring the suite through a mechanism that permanently overrides the app under hash routing. No library change, hence no new package unit tests.

## Fix

`LOCATION_PLUGIN` now seeds `sessionStorage.featureFlags` (the app's own persistence layer, priority 2 in `resolveLocationPluginFeature()`) via `cy.visit` `onBeforeLoad`, instead of the query string. Explicit per-spec `features` still go through the query param — `location_plugins.cy.js` exercises that URL mechanism itself and keeps doing so.

## CI legs

- `sample-app-lit-e2e`'s `test:cypress:all` now runs **all five suites concurrently** (vanilla, mobx, docs, hash, navigation) against the one wrangler server; the `test` script keeps its main-side form. It rides the existing `ci` umbrella task; `build-test.yml` is unchanged from `main` (the earlier bespoke **E2E (hash mode)** step and `test:hash` turbo task are gone, per review). Isolation is sound — each `cypress run` is its own browser process/profile and the suite-wide plugin is seeded per-visit via `onBeforeLoad`. Honest caveat: CI gives 4 vCPUs to 5 Electron processes. Measured on this PR's runs: per-suite time inflates under the contention (vanilla 38 s sequential → 61 s five-way) but the e2e wall drops (~73 s → ~61 s), and both build_and_test runs on the parallel head were green. Locally: 1m00 five-way vs 1m18 batched, 5/5 green. If the oversubscription starts flaking, the sequential two-batch variant (438323d) is the proven fallback.
- The navigation batch member extends coverage to the Navigation API plugin — the package's flagship mode — so all three published location-plugin configurations (`pushState`, `hash`, `navigation`) run on every CI pass.
- `test:hash` / `test:navigation` remain as standalone scripts (each starts its own server) for running a single mode locally: `pnpm --filter sample-app-lit-e2e run test:hash`.

## Evidence (local, wrangler-served `docs/dist`)

| run | before (main) | after |
| --- | --- | --- |
| vanilla, hash | 25/27 — `feature_flags_panel.cy.js` 0/2 | **27/27** |
| vanilla, pushState | — | 27/27 |
| mobx, pushState | — | 27/27 |
| vanilla, navigation | — | 27/27 |
| `turbo run test` (final shape: five concurrent suites) | — | 5/5 + 27/27 ×4, 1m00 local |


.github/workflows/build-test.yml
```diff
@@ -76,6 +76,15 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
           # Silence per-request [wrangler:info] logging from the e2e dev server.
           WRANGLER_LOG: warn
+      - name: E2E (hash mode)
+        # The main run's Cypress suites cover pushState only; hash routing is a
+        # published navigation-location-plugin configuration.
+        run: turbo run test:hash --ui=stream --log-order=stream ${{ inputs.force && '--force' || '' }}
```
```suggestion
      - name: E2E (hash mode)
        run: turbo run test:hash --ui=stream --log-order=stream ${{ inputs.force && '--force' || '' }}
```

simshanith: tho really why can't this be rolled into the ci task?

simshanith: Rolled in via 438323d — the test script now runs the vanilla/mobx/docs batch and then a hash+navigation batch against the same server, so the leg rides `turbo run ci` and the bespoke workflow step (this suggestion's target) is deleted along with the `test:hash` turbo task. The navigation batch member also extends coverage to the Navigation API plugin per follow-up ask. Verified locally: all five suites green in one `turbo run test` (docs 5/5, vanilla/mobx/hash/navigation 27/27 each).

apps/sample-app-lit-e2e/package.json
```diff
@@ -10,11 +10,17 @@
     "format": "oxfmt \"**/*.{ts,js,json,md}\"",
     "format:check": "oxfmt --check \"**/*.{ts,js,json,md}\"",
     "lint": "oxlint .",
-    "test": "start-server-and-test docs:start-server 'http://localhost:8787/app/|http://localhost:8787/app-mobx/' test:cypress:all",
+    "test": "start-server-and-test docs:start-server 'http://localhost:8787/app/|http://localhost:8787/app-mobx/' test:cypress:full",
     "test:cypress": "cypress run",
     "test:cypress:all": "concurrently --names vanilla,mobx,docs --prefix-colors auto \"pnpm run test:cypress\" \"pnpm run test:cypress:mobx\" \"pnpm run test:cypress:docs\"",
     "test:cypress:docs": "cypress run --config-file cypress.docs.config.ts",
-    "test:cypress:mobx": "cypress run --config baseUrl=http://localhost:8787/app-mobx/,videosFolder=cypress/videos-mobx,screenshotsFolder=cypress/screenshots-mobx"
+    "test:cypress:full": "pnpm run test:cypress:all && pnpm run test:cypress:modes",
```
simshanith: seems like `test:cypress:modes` should fold into `test:cypress:all` and that + this can go away+revert line 13

simshanith: Done in f9297df — `test:cypress:modes` folded into `test:cypress:all` (five-way `concurrently`), the intermediate `test:cypress:full` deleted, and the `test` script reverted to its main-side form. Locally 5/5 suites green in 1m00 vs 1m18 for the sequential batches. Note for CI: that's 5 Electron processes on a 4-vCPU runner — isolation is fine (per-process browser profiles, per-visit sessionStorage seeding), CPU oversubscription is the only watch item.

PR #294, opened 2026-07-11
ci(tooling): commit-msg hook runs commitlint locally

simshanith:
Was stacked on #269 (now merged; auto-retargeted to `main`). This is the standing precondition for #269's CI check: contributors get commit-time feedback locally before `lint_pr_commits` ever runs, using the same `commitlint.config.mjs` (config-conventional + the hand-typed-merge ignore).

## Tool choice

husky, simple-git-hooks, and bare `core.hooksPath` all reduce to the same mechanism (point git at a tracked hooks dir on install), so the tie-breaker is the ecosystem standard: husky is what commitlint's own docs document (`prepare` script + `.husky/commit-msg`), it's dependency-free, and its footprint here is three lines of config plus a one-line hook file. simple-git-hooks would inline the command into package.json but is the less-traveled path; bare `core.hooksPath` needs a custom install script that ends up reimplementing husky with no differentiator. No repo-specific reason to deviate was found, so this defaults to the standard.

## CI: `HUSKY: 0` everywhere, load-bearing in one place

Per husky's own CI recommendation, every workflow that runs `pnpm install` (build-test, bump-version, publish-gh, publish-npm) now sets workflow-level `env: HUSKY: 0`, skipping the pointless hook install on CI. On **bump-version it is required, not conventional**: release-it commits `Release X.Y.Z` — a message the commit-msg hook rejects by design (the release exemption is a label on the GitHub check, not in the config). Verified both halves empirically:

- installed release-it 20.2.1 defaults `git.commitArgs: []` and its `Git.commit()` adds no `--no-verify`, so the hook would fire on the release commit;
- with hooks installed, `git commit -m "Release 9.9.9"` → rejected (subject-empty, type-empty, exit 1); `HUSKY=0 git commit` and `git commit -n` → pass.

publish-gh pushes tags only and publish-npm commits nothing (`git checkout --` restore only), so for those and build-test the env is purely the standard skip. `.husky/_` shims for hooks without a user script exit 0, so tag pushes are unaffected either way.

## pnpm 11 caveat

pnpm 11's "Already up to date" fast path skips lifecycle scripts, so `prepare` won't fire on no-op installs — verified: after deleting `.husky/_`, a no-op `corepack pnpm install` left it missing, and `corepack pnpm run prepare` restored it. CONTRIBUTING documents that escape hatch plus a link to [husky's how-to](https://typicode.github.io/husky/how-to.html) for the per-commit (`git commit -n`) and env (`HUSKY=0`) bypasses.

## Verification (in a worktree)

- Fresh `corepack pnpm install` → `core.hooksPath` = `.husky/_`, shims present.
- `git commit -m "Update stuff"` on a scratch file → **rejected** (subject-empty, type-empty; `husky - commit-msg script failed (code 1)`).
- `chore: scratch` → passes. `fixup! chore: scratch` (default ignore) → passes. `Release 9.9.9` → rejected; passes only with `HUSKY=0` or `-n`. Scratch commits reset away.
- All branch commits lint clean via `commitlint --from` over the branch range.
- `turbo run lint format:check test:root typecheck:root` → 42/42 green; `actionlint` clean on the workflow edits.


.github/workflows/build-test.yml
```diff
@@ -20,6 +20,10 @@ on:
         default: false
 
 # Cancel superseded runs per event per ref; never cancel main runs.
+# Standard husky CI setup: skip the prepare-script git-hook install on CI installs.
+env:
+  HUSKY: 0
+
```
simshanith, line 25: this interrupts the concurrency comment

PR #295, opened 2026-07-11
refactor(tooling): typecheck the commitlint config

simshanith:
Refs #269 (which introduced the config) and #13.

## Motivation: the config was statically invisible

Empirical probes on main showed `commitlint.config.mjs` escaped every static layer:

- **oxlint**: the blanket `**/*.{js,mjs,cjs}` override switches off most rules — an injected `const unused = 1` produced zero findings.
- **oxfmt**: `format:check:root`'s glob `*.{md,ts,js,json,jsonc,yml,yaml}` doesn't match `.mjs` at all.
- **TypeScript**: in no tsconfig — and force-feeding a tsconfig-orphaned file to the type-aware layer is the known fake-clean/panic trap.

Its only guard was functional (the hook and `lint_pr_commits` load it). This PR makes it a first-class `.ts` file and adds the enforcement arm so no future `.mjs`/`.cjs` file slips back into that hole.

## What changed (one commit each)

1. **`commitlint.config.mjs` → `commitlint.config.ts`** — typed via `import type { UserConfig } from '@commitlint/types'` + `satisfies UserConfig` (new `@commitlint/types` catalog devDep); added to the root tsconfig `include`; oxlint `ignorePatterns` gets `!**/commitlint.config.ts` following the `!**/vitest.config.ts` precedent; CONTRIBUTING reference updated. The workflow now passes `configFile: commitlint.config.ts` explicitly — wagoid's default is the old `.mjs` name and it **silently falls back to bare config-conventional** (dropping our custom ignore) when that file is missing.
2. **Root fmt glob** gains `mjs,cjs` (and the turbo `format`/`format:check` inputs over-approximate to match), so any future root `.mjs`/`.cjs` is at least format-checked.
3. **`check:mjs` guard** (`scripts/check-mjs.{core,test,}.ts`, check-catalog pattern): `git ls-files '*.mjs' '*.cjs'` (tracked files only — node_modules/dist can never enter) diffed against `MJS_ALLOWLIST` (path → one-line reason, **seeded empty** — the config was the last tracked `.mjs` in the repo). Unlisted files fail with the two options (convert to `.ts`, or allowlist with a reason); stale entries fail too, so the list stays short. Wired into the `ci` pipeline beside `check:catalog`/`check:pack`. Every allowlisted file is a documented static-analysis hole — the empty list is the point.
4. **Lint wins for plain-JS files**: the `**/*.{js,mjs,cjs}` override dates from the eslint→oxlint migration (#231), when the repo carried many `.mjs` scripts; it switched off four purely syntactic rules alongside the type-aware ones. Re-enabled: `no-redeclare`, `no-prototype-builtins`, `no-unused-vars`, `no-unused-expressions` — and fixed the 12 real findings they exposed in `apps/sample-app-shared`'s legacy `.js` files (unused vars/params, `var` redeclarations, direct `hasOwnProperty`, a sequence-expression statement in the GA bootstrap). Stays off, with reasons: `typescript/no-require-imports` (the surviving CJS-flavored fixture generator exists precisely to use `require`; converting it is a behavior change, not a lint fix) and all `typescript/*` type-aware rules (these files belong to no tsconfig — tsgolint cannot run there).

## Consumer verification

| Consumer | Loader | Result |
|---|---|---|
| Local CLI (`echo "feat: x" \| pnpm exec commitlint`) | commitlint v21 + jiti | ✅ good passes, bad fails, and the custom lowercase-merge ignore passes — which **proves the `.ts` file loaded** (a fallback config wouldn't have it) |
| Hook path (`pnpm exec commitlint --edit FILE`, #294's mechanism) | same | ✅ good passes, bad fails |
| CI (`wagoid/commitlint-github-action@v6`) | bundles `@commitlint/load@^19.2` → `cosmiconfig-typescript-loader@^5` → hard dep on `jiti` | ✅ by evidence: the action's source rejects only `.js` (`.js extension is not allowed … use .mjs instead`) and hands any other existing path to `load({}, { file })`; **this PR's own `lint_pr_commits` run is the definitive empirical test** — it self-validates like #269 did |

## Static coverage, before → after

| Layer | before (.mjs) | after (.ts) |
|---|---|---|
| oxlint | neutered override; injected junk: 0 findings | full ruleset; injected `const unused = 1` → `no-unused-vars` error (verified, reverted) |
| oxfmt | not matched by any glob | matched; injected mis-format flagged (verified, reverted) |
| typecheck | no tsconfig | root tsconfig; wrong `ignores` type → TS2322 (verified, reverted). Note: `UserConfig` tolerates unknown keys, so this checks known-key shapes, not excess keys |
| regression guard | none | `check:mjs` fails on any new tracked `.mjs`/`.cjs` (probe: staged dummy file → named in failure with both options; verified, reverted) |

## Verification

- `turbo run lint format:check test:root typecheck:root check:catalog check:pack check:mjs`: 45/45 tasks (repo-wide lint including the re-enabled rules)
- `lint:workflows` (actionlint) clean; `lint:package-json` 16/16
- All four commit messages lint clean (they're the squash-body bullets)

## Interaction with #294

The commit-msg hook PR invokes `commitlint --edit`, which is config-filename-agnostic (verified above), and its diff contains no `commitlint.config.mjs` references — merge order is flexible; whoever lands second rebases trivially.


.github/workflows/commitlint.yml
```diff
@@ -36,6 +36,10 @@ jobs:
         with:
           # Full history so the action can walk the PR's commit range
           fetch-depth: 0
-      # Lints with the repo's commitlint.config.mjs (merge/revert/fixup
-      # exemptions live there; the action loads only ESM .mjs configs)
+      # Lints with the repo's commitlint config (merge/revert/fixup
+      # exemptions live there). configFile must be explicit: the action's
+      # default is commitlint.config.mjs and it silently falls back to bare
+      # config-conventional when that file is missing.
```
simshanith:
```suggestion
```

simshanith: Applied in a2e741f — comment block dropped, `configFile: commitlint.config.ts` stays.

PR #300, opened 2026-07-12
feat(tooling): eslint-plugin-pnpm replaces check:catalog + oxlint overlap suppression

simshanith:
## Current state (before this PR)

- ESLint 10 flat config at root `eslint.config.js`, retained solely for **package.json linting**: `eslint-plugin-package-json` recommended preset (jsonc parser comes with it) plus `package-json/require-description` with `ignorePrivate`.
- Entry point: root script `lint:package-json` = `eslint --format tap "**/package.json"` → turbo `//#lint:package-json` → attached `with` every package `lint` task → `ci`. CI runs `turbo run ci` (build-test.yml).
- **No JS/TS file is eslint-linted** — oxlint owns all code linting (`.oxlintrc.json`, `lint:root` + per-package `lint`).
- Note: `eslint-plugin-turbo` is a root devDep consumed by **oxlint**, not by `eslint.config.js` — `.oxlintrc.json` loads it via `jsPlugins: ["eslint-plugin-turbo"]` and uses `turbo/no-undeclared-env-vars` (rules + three overrides).

## What this PR does

1. Adopts **eslint-plugin-pnpm 1.6.1** for `package.json` *and* `pnpm-workspace.yaml` (script now passes `pnpm-workspace.yaml` too; turbo inputs updated).
2. Hoists the previously-legal single-consumer inline deps into the catalog so `json-enforce-catalog` is green — this *tightens* policy from "no inline dep in 2+ packages" (check:catalog) to "no inline dep at all" in workspace manifests. The hoist was produced by the rule's own autofix and re-verified: `pnpm install` clean, lockfile resolutions unchanged.
3. Adds **eslint-plugin-oxlint 1.73.0** as the last config entry via `buildFromOxlintConfigFile('./.oxlintrc.json')` (the config-file-aware variant, since the repo curates a large `.oxlintrc.json`).
4. Moves vue-check's deliberate real-TS6 pin into a new **`typescript6` named catalog** (`typescript: ^6.0.3`). Per #257, vue-tsc cannot use `catalog:typescript6-compat`: Volar resolves `typescript/lib/tsc` as a *path*, and the `@typescript/typescript6` alias only re-exports that file. Same bits installed (verified: `tools/vue-check/node_modules/typescript` is real `typescript 6.0.3`; docs typecheck via vue-tsc green). This lets the rule's `ignores: ['typescript']` carve-out be **deleted** — inline `typescript` is now flagged like everything else (probe-verified).
5. Moves dts-backtest's aliased consumer-TS matrix (`typescript-5.0`…`typescript-7`, `npm:typescript@*` specs) into the catalog — pnpm 11 catalogs hold `npm:` alias specs fine (verified: each alias resolves to real `typescript` at its pinned floor). Per review, the 6.x leg shares the `typescript6` named catalog with vue-check (catalog entries key by dep name — `catalog:typescript6` on the alias name errors with `ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC` unless the catalog holds a `typescript-6` entry — so the alias entry moved in next to vue-check's; bumping that catalog bumps both consumers deliberately, noted in dts-backtest's README). With that, `npm` drops out of `allowedProtocols`: inline `npm:` aliases now error too (probe-verified).
6. **Retires `check:catalog`**: deletes `scripts/check-catalog{,.core,.test}.ts`, the root `check:catalog` script, and the turbo `//#check:catalog` task (ci `dependsOn` updated). `scripts/types.ts` and `scripts/workspace.ts` stay — `check:pack` still imports them. `test:root` glob still matches the remaining `check-pack`/`start-xvfb` suites (13 tests green).

## Rule adoption table (each adopted rule probe-verified: violated → error observed → reverted)

| Rule | Decision | Config | Probe evidence |
|---|---|---|---|
| `json-enforce-catalog` | **adopt** | `allowedProtocols: [workspace, link, file, portal]` — tighter than check:catalog's managed prefixes: `npm:` aliases are catalog-expressible (dts-backtest's matrix moved there), so they are no longer exempt; `fields`: all 4 dep fields (same as the script); `conflicts: 'error'` (see plugin-bug note below); no `ignores` — vue-check's TS6 pin moved to the `typescript6` catalog instead; off for `examples/*/package.json` (standalone `npm ci`/StackBlitz projects need real versions) | `^0.11.0` hardcoded where `catalog:` belongs → `Expect to use catalog instead of plain specifier, got "^0.11.0" for package "@custom-elements-manifest/analyzer"` |
| `json-valid-catalog` | **adopt** | `autoInsert: false` (default fix invents `^0.0.0` catalog entries on `--fix`) | `"vue": "catalog:doesnotexist"` → `Catalog "catalog:doesnotexist" for package "vue" is not defined in pnpm-workspace.yaml` |
| `json-prefer-workspace-settings` | **adopt** | defaults | `pnpm.overrides` added to root package.json → `Unexpected pnpm settings in package.json, should move to pnpm-workspace.yaml` |
| `yaml-no-unused-catalog-item` | **adopt** | defaults | unused `left-pad: ^1.3.0` catalog entry → `Catalog item "left-pad:default" is not used in any package.json` |
| `yaml-no-duplicate-catalog-item` | **adopt** | `checkDuplicates: 'exact-version'` — the default `name-only` flags all 6 *intentional* cross-catalog pins (`@uirouter/core`, `lit`, `mobx`, `typedoc` in `publishedPeer`; `typescript` in `typescript6-compat`; `vite-plugin-static-copy` in `vitepress1`), verified by running the default against unmodified pnpm-workspace.yaml | verbatim `vite: ^8.1.3` duplicated into a named catalog → `Catalog item "vite" with version "^8.1.3" is already defined…` |
| `yaml-valid-packages` | **adopt** | defaults | `- missing-dir/*` added to `packages:` → `Package pattern "missing-dir/*" does not match any directories with a package.json file` |
| `yaml-enforce-settings` | **reject** | — | opt-in policy rule (you list exact settings to pin); nothing to enforce today, oxfmt/reviews cover the file |

Known quirk found while probing `yaml-no-duplicate-catalog-item`: it only compares against the *first* occurrence of a name, so with `exact-version` a duplicate between the default catalog and a second named catalog is missed when a first named catalog already holds that name at a different version. Acceptable: it still catches the common copy-paste case.

## check:catalog retirement: behavior-by-behavior equivalence (probed before deleting)

What the script enforced: any dep declared with an inline registry range (managed prefixes exempt) in **2+** workspace packages, across dependencies/devDependencies/peerDependencies/optionalDependencies, must be hoisted to the catalog; URL/git specs exempt; reports drift and fix hints.

| check-catalog.core.ts behavior | Plugin coverage | Probe |
|---|---|---|
| Inline range in `dependencies` → violation site | `json-enforce-catalog` | inline `screenfull ^6.0.2` in docs → error |
| … in `devDependencies` | same | inline `@custom-elements-manifest/analyzer ^0.11.0` → error |
| … in `peerDependencies` | same (`fields`) | inline `mobx ^6.0.0` peer in lit-ui-router-mobx → error |
| … in `optionalDependencies` | same (`fields`) | inline `left-pad ^1.3.0` optional dep → error |
| 2+ distinct consumers required | **stricter: flags at 1+** | `lodash ^4.17.0` + `^4.18.0` in two apps → both sites error |
| Version-drift report across consumers | subsumed — every inline site errors individually | same lodash probe shows both drifting specs |
| Managed-prefix exemptions (`catalog:`/`workspace:`/`link:`/`file:`/`portal:`/`npm:`) | `allowedProtocols` + native catalog/workspace handling; **stricter on `npm:`** — aliases moved to the catalog and inline ones now error | dts-backtest matrix green via `catalog:`; probe: inline `npm:typescript@5.0.4` → error |
| URL/git specs exempt (`://`, `git+`, `github:`, …) | **stricter: plugin flags them** | `github:left-pad/left-pad` → error. Not a gap — errs toward more discipline; a git dep would go in the catalog (pnpm allows git specs there) or get an `allowedProtocols` entry |
| Workspace-member enumeration via pnpm SDK (14 members) | eslint glob lints all 17 tracked package.json files; `examples/*` (never workspace members) exempted from enforce-catalog only | green run covers the full glob |
| "already in catalog" / "add to catalog" fix hints | autofix performs the edit itself | the hoist commit in this PR was produced by the rule's fixer |
| Former `ignores: ['typescript']` escape hatch | **closed** — carve-out deleted after the `typescript6` catalog move | inline `typescript ^6.0.3` in vue-check → error |

Nothing the script checked is left uncovered; the plugin is equal-or-stricter on every behavior, so the script, its tests, and its turbo task are deleted in this PR.

**Plugin bug found while probing:** with the default `conflicts: 'new-catalog'`, a *plain no-`--fix`* eslint run of `json-enforce-catalog` **writes to pnpm-workspace.yaml** when an inline spec conflicts with an existing catalog entry (reproduced: inline `typescript ^6.0.3` vs catalog `^7.0.2` spawned a `conflicts_typescript6-compat_h6_0_3` catalog; verified via checksum before/after). `conflicts: 'error'` stops the mutation while still reporting the violation — hence that setting. Worth an upstream issue.

## eslint-plugin-oxlint: honest no-op disclosure

Today ESLint lints only `package.json` files and `pnpm-workspace.yaml`, so the oxlint suppression config **currently disables nothing** — there are no JS/TS files in the eslint run for oxlint-covered rules to apply to. It's wired anyway because the cost is one config line + one catalog dep, and it future-proofs any eslint expansion (the moment a JS/TS glob is added, oxlint-covered rules start suppressed instead of double-reported). Verified empirically: `buildFromOxlintConfigFile('./.oxlintrc.json')` tolerates the config's `jsPlugins`/`turbo/*` entries gracefully — it parses the file into 12 flat configs (132 rule suppressions) and silently ignores the external-plugin fields, no error or warning. Friction to know about: its peer range is `oxlint: ~1.73.0`, so future oxlint catalog bumps need a matching eslint-plugin-oxlint bump.

## Install / catalog notes

- Both plugins installed clean under pnpm 11 `minimumReleaseAge` (1.6.1 is from 2026-05-19, 1.73.0 from 2026-07-07); no auto-written exclusions.
- New deps added as `catalog:` refs, alphabetized; the hoist also moves the pre-existing root eslint stack (`eslint`, `eslint-formatter-tap`, `eslint-plugin-package-json`, `eslint-plugin-turbo`, `pnpm`, `turbo`) into the catalog.

## Verification

- `turbo run ci` foreground: **67/67 tasks green** (was 68/68 before the retirement — the delta is exactly the removed `//#check:catalog` task)
- `pnpm run lint:package-json`: green, 18 files (17 package.json + pnpm-workspace.yaml)
- `check:catalog` ran green right up to its deletion (14 workspace packages, no violations)
- `test:root` 13 tests green (check-pack + start-xvfb suites still satisfy the `scripts/**/*.test.ts` glob); `lint:root`, `format:check:root`, `pnpm install --frozen-lockfile` green
- docs typecheck (vue-tsc through the new `typescript6` catalog) green
- dts-backtest floor intact through the catalog move — per-leg resolved versions self-reported by the tool: TS 5.0.4 / 5.9.3 / 6.0.3 / 7.0.2 each check both fixture tsconfigs green, and the NoInfer selftest confirms the 5.0.4 floor leg really is old TS (`floor TS 5.0.4 rejects the probe, current TS 6.0.3 accepts it`; native 7.0.2 rejects)
- Probe evidence above; workflows untouched (no actionlint run needed)


pnpm-workspace.yaml
```diff
@@ -57,6 +71,9 @@ catalogs:
     typedoc: ^0.28.0
   typescript6-compat:
     typescript: npm:@typescript/typescript6@^6.0.2
+  # real typescript 6 for vue-tsc: Volar resolves typescript/lib/tsc as a path the compat alias only re-exports
```
```suggestion
```

pnpm-workspace.yaml
```diff
@@ -35,16 +42,28 @@ catalog:
...
   release-it: ^20.2.1
   rimraf: ^6.1.3
+  screenfull: ^6.0.2
+  start-server-and-test: ^3.0.11
+  turbo: ^2.10.4
   typedoc: ^0.28.20
   typedoc-plugin-markdown: ^4.12.0
   typedoc-vitepress-theme: ^1.1.3
   typescript: ^7.0.2
+  # dts-backtest's consumer-TS matrix: aliased real typescript at each floor
```
```suggestion
```

PR #339, opened 2026-07-12
feat(tooling): monorepo mise tasks for the release pipeline

simshanith:
Converts the nine `release_*` tasks from root-config prefix naming to mise monorepo task addressing: the task defs now live beside the code they run, in `tools/release/mise.toml`, addressed as `//tools/release:<step>`. This resolves the root config reaching into tools/release — the run lines are config_root-relative (`node src/steps/…`), matching #338's layout. Stacked on `chore/zizmor` (#331, which absorbed #338).

## Why monorepo addressing won (vs the `dir = "tools/release"` fallback, vs `pnpm --filter exec`)

All viability gates passed empirically on the pinned mise 2026.6.14:

| Concern | Result |
| --- | --- |
| Experimental flag | **Not needed.** `monorepo_root = true` + `[monorepo].config_roots` is the stable, current shape (`experimental=false` throughout). Filesystem auto-discovery is what's deprecated — explicit `config_roots` avoids it. |
| Trust | **No new burden.** Trusting the monorepo root covers every descendant config_root (mise writes a `.monorepo` trust marker). Verified: fresh worktree + single `mise trust --quiet` runs `//tools/release:*` with zero extra prompts; an explicit `mise trust --untrust` of the nested file is overridden by root trust. |
| CI / mise-action | jdx/mise-action sets `MISE_TRUSTED_CONFIG_PATHS=$PWD`; verified that env var also covers nested config_roots. No workflow changes beyond the task names. |
| Root tasks | `mise run setup` / `lint_workflows` unprefixed at root keep working; workflows stay single-line `mise run` steps. |
| Inheritance | Root `[env]` `_.path` and `[tools]` (node via .nvmrc, npm, corepack) reach nested tasks; `depends = ["//:setup"]` crosses configs and runs the same corepack→setup graph. |
| Lockfile | No `tools/release/mise.lock` appears; root mise.lock untouched. (mise's monorepo per-subproject lockfile default only changes in 2027.6.0, and the nested config is [tasks]-only.) |

The `dir` fallback also works (probed: `dir` is config_root-relative) but keeps all nine defs in the root config — placement, not locality. `pnpm --filter exec` adds a stdout/exit-code layer for nothing; never built.

## Shape

- Root `.config/mise/config.toml`: `monorepo_root = true`, `[monorepo] config_roots = ["tools/*"]` (glob only matches dirs that carry a mise config — `tools/build_and_test` et al. can join later); provisioning + lint tasks stay; release section is now a pointer comment.
- `tools/release/mise.toml`: the nine tasks, `release_` prefix dropped in favor of path addressing (`//tools/release:pack` etc.), env-contract comments preserved, **[tasks]-only** (kept inert for shells that cd in; root env/tools inherited).
- One behavioral pin: `tasks.build` sets `dir = "../.."` — turbo run from inside a package silently adds that package to the scope alongside the explicit `<pkg>#build` spec (observed: "Packages in scope: release, ui-router-navigation-location-plugin"); pinning the cwd restores the exact previous scope.
- Workflows: `mise run release_pack` → `mise run //tools/release:pack` (still single-line; `//…:name` needs no shell quoting).

## Gates

- All nine tasks ran secret-free from repo root **before and after** the conversion; exit codes identical (`git_user` refuses outside Actions; `tag`/`publish`/`bump` reach the same interactive release-it prompt with byte-identical argv; `package_info`/`build`/`pack`/`check_tarball` green with identical GITHUB_OUTPUT).
- `mise run lint_workflows` (actionlint + zizmor --offline): no findings.
- Fresh-worktree bootstrap probe: `mise trust --quiet && mise install && CI=true mise run //tools/release:package_info` — frozen-lockfile install, correct outputs, no trust prompts.
- `turbo run lint typecheck test --filter=release`: 8/8 green.
- `mise tasks ls --all` lists all 16 tasks (`//:` root + `//tools/release:` nine).

One discovery-cost note: bare `mise tasks ls` at the repo root lists only root tasks; the nested nine need `--all` (or any subdir cwd). The root config's pointer comment says so.

### DIFF
--- a/.config/mise/config.toml
+++ b/.config/mise/config.toml
@@ -1,4 +1,15 @@
 # One version authority per tool — see the toolchain table in CONTRIBUTING.md.
+
+# Monorepo task addressing (`mise run '//tools/release:pack'`): task defs live
+# beside the code they run. Root trust covers every config_root below, so the
+# usual `mise trust` (or mise-action's MISE_TRUSTED_CONFIG_PATHS) is enough.
+monorepo_root = true
+
+[monorepo]
+# Explicit roots — mise deprecated filesystem auto-discovery. A glob only
+# matches directories that actually carry a mise config.
+config_roots = ["tools/*"]
+
 [env]
 # Workspace binaries win over mise tools/globals; absolute via config_root
 # (repo root), so bare `turbo` etc. stay spawn-safe in any cwd or worktree.
@@ -23,8 +34,8 @@ idiomatic_version_file_enable_tools = ["node"]
 
 # Provisioning tasks, not hooks: hooks warn-and-continue, task failures are
 # loud; and tool-only bumps (actionlint) skip the corepack ceremony.
-# snake_case names (bazel target convention, where mise monorepo tasks are
-# trending); no colons — mise monorepo task addressing claims `:`.
+# snake_case names (bazel target convention); no colons in task names — mise
+# monorepo task addressing owns `:` (see the config_roots above).
 [tasks.corepack_enable]
 description = "Shim the corepack-managed package managers onto PATH"
 run = "corepack enable"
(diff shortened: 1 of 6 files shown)


.config/mise/config.toml
```diff
@@ -64,66 +75,6 @@ description = "Lint and audit GitHub Actions workflows"
...
-depends = ["setup"]
-run = "node tools/release/src/steps/release-publish.ts"
-
-[tasks.release_bump]
-# env: PACKAGE, INCREMENT, OTHER_INCREMENT, PR_BASE, BRANCH_PREFIX_INPUT,
-# DRY_RUN, GH_TOKEN (gh pr create; the branch push rides checkout's PAT).
-description = "Compute the bumped version, branch, commit, push, and open the release PR"
-depends = ["setup"]
-run = "node tools/release/src/steps/release-bump.ts"
+# The nine release-workflow tasks live beside their code in
```
```suggestion
# The release-workflow tasks live beside their code in
```

simshanith: Applied in 0c04f66.

PR #369, opened 2026-07-17
build(lit-ui-router): pass-split build via @tools/oxc-emit — oxc JS + isolated-declarations d.ts + shipped source maps

simshanith:
Converts lit-ui-router's build from single-pass `tsc` to the repo's new **pass-split build convention**, with both passes emitted by oxc via a graduated tool — and ships working source maps for the first time.

## `@tools/oxc-emit` — two dependency-free passes
- **`oxc-emit-js`**: oxc `transformSync` (legacy-decorator lowering with tsconfig.base parity: `decorator.legacy` + `setPublicClassFields` + `removeClassFieldsWithoutInitializer` ≙ `experimentalDecorators` + `useDefineForClassFields: false`) → codegen-only `oxc-minify` (`compress:false, mangle:false`) so **comments never ship** — comment edits are publish-neutral under `check:published-diff` forever.
- **`oxc-emit-dts`**: oxc **isolated declarations** — syntax-derived d.ts with JSDoc intact. No typechecking, no cross-package inputs.

Both passes take milliseconds and depend on nothing, so every pass of every converted package can run concurrently. Sources walk via stdlib `fs.globSync` (pattern-array `exclude`, no warning on pinned Node 24).

## Source maps, both passes
- **JS**: `sourcemap: true` through both stages, composed via `@ampproject/remapping` (`excludeContent: true`) into one `dist/*.js.map` — mappings survive the comment-stripping stage and land on real src lines.
- **d.ts**: `isolatedDeclarationSync(..., { sourcemap: true })` → `dist/*.d.ts.map`. Net-new capability — the old tsc build never shipped `declarationMap`; consumer go-to-definition now lands on actual sources.
- `files: ["dist/**", "src/*.ts", "!src/*.spec.ts"]` ships the 12 sources as map referents (specs verified absent from the pack). No `sourcesContent` anywhere: maps stay lean, comments live only in shipped `src/**`.

## Root convention (turbo.json)
- `build` is an **umbrella**: `dependsOn: ["^build", "build:js", "build:types"]` — own-package pass deps, transit no-ops where the scripts don't exist, so `^build` keeps meaning "everything a dependency ships". No package defines a `build` script.
- `build:js`: no deps. `build:types`: `^build:types` as **ordering propagation** for transitive consumers (not a compile input — isolated declarations need no upstream d.ts).
- `typecheck`: `["^build", "^build:types"]` — `^build` stays while single-pass packages exist. `typecheck:src` rides `typecheck`'s `with` (no ordering edge, fully parallel).
- cem is **not** root vocabulary (single-package concern): lit-ui-router's turbo.json overrides `build` with only `"with": ["build:custom-elements"]` — per-key merge keeps the root `dependsOn` (verified via `--dry=json`).

## isolatedDeclarations, enforced
`tsconfig.src.json` (renamed from tsconfig.build.json — nothing builds from it; it names the src conformance surface) sets `isolatedDeclarations: true`; `typecheck:src` runs it, scope-suffixed like the existing `typecheck:tests` pattern, while plain `typecheck` covers everything with specs exempt. Cost: **40 mechanical annotations** on the public surface. mobx and nav-plugin owe 4 each; ui-router-server already conforms.

## `check:published-diff`: ship-affecting vs ship-inert
Drift now classifies via `isShipAffecting`: **ship-affecting** (`dist` js/d.ts + manifest) drives the release-owed verdict; **ship-inert** (`src/**`, `dist/**/*.map`) is printed with hollow `◦ (ship-inert)` bullets but never owes a release — atomic packs make wrong-line mapping impossible, and any real code change also changes the emitted js. Against published 1.7.0: **27 ship-affecting files, 37 ship-inert**; mobx + nav-plugin stay clean.

## Published-artifact changes vs 1.7.0
1. All `dist/*.js`: comments stripped, oxc style, `_decorate` helper imported from `@oxc-project/runtime` (new runtime dependency, pinned lockstep with the oxc trio — same model as tsc `importHelpers` + `tslib`).
2. All `dist/*.d.ts`: oxc isolated-declarations emit (JSDoc intact; dts-backtest holds the TS 5.0 consumer floor).
3. Maps: the old broken `dist/*.js.map` (no `sourcesContent`, sources → unshipped paths) are replaced by working composed js maps + net-new d.ts maps, with `src/*.ts` shipped as referents.
4. `package.json`: `@oxc-project/runtime` dependency + the `files` field.

dist: 252K → ~150K unpacked, ~38K → ~23K gzipped (npm install weight — comment/JSDoc stripping).

## Bundle-size attribution (codecov: −656 B/app, all in the directive chunk)
Measured cause: **two inline `__decorate` helpers eliminated** (~−560 B) — tsc inlined per decorated file; oxc imports the identical TypeScript-authored helper from `@oxc-project/runtime`, which dedupes into the pre-existing shared chunk (byte-identical before/after, so the new dep costs ~0 in bundles). Remainder: equivalent enum lowering (−38 B). Comment stripping contributes zero to app bundles (minification strips both sides equally); three of six modules minify byte-identical old-vs-new. Purely bytes, no functional change; coverage identical.

## Ride-along fix: vitest `@types/node` peer-keying
Removing lit-ui-router's `@types/node` (a leftover from the early in-package build script) exposed a latent bug: vitest optional-peers on `@types/node`, and the packages that never declared it (mobx, nav-plugin) were getting pnpm's auto-peer at **latest (25.0.9)** — outside the catalog pin — splitting the workspace into two peer-keyed vitest type-instances that happened to be internally coherent. The fix declares `@types/node: catalog:` in **every** package with a `vitest.config.ts` (all three read `process.env`, so each declaration is true): one catalog-pinned vitest instance, phantom `@types/node@25` cluster gone (−236 lockfile lines), and lit-ui-router's line survives on honest grounds.

## Verification
Full `turbo run ci` green — 95 tasks: dts-backtest on the oxc d.ts, sample-app cypress e2e against the oxc dist, per-key-merge and transit-node graph shapes via `--dry=json`, tools/release suite (102 tests incl. the ship-inert split), map traces (`dist/core.js 62:20 → ../src/core.ts 253:22`; `dist/core.d.ts → litViewsBuilder`), byte-identical dist across the globSync refactor.

## Follow-ups (sequenced)
mobx/nav-plugin adoption PRs (4 annotations each + their tsconfig renames); then drop `^build` from root `typecheck` once all packages are pass-split.

Ship-affecting — the natural 1.7.1 vehicle (lands when review says so). Maps work was reviewed as stacked PR #370, folded in here.


packages/lit-ui-router/turbo.json
```diff
@@ -3,7 +3,8 @@
   "extends": ["//"],
   "tasks": {
     "build": {
-      "dependsOn": ["build:custom-elements"]
+      // per-key merge: dependsOn/outputs inherit from the root umbrella
```
simshanith:
```suggestion
```

PR #424, opened 2026-07-19
build: bump vitest to v5 and drop the @vitest/browser patch

simshanith:
> [!IMPORTANT]
> **STILL PARKED — but the reason has changed.** The original exit condition (vitest-dev/vitest#9381 shipping in a release) is **MET**: it first shipped in `vitest@5.0.0-rc.1`, the bump is on this branch (now tracking **`5.0.0-rc.2`**), and the patch is gone. The only remaining hold is that **5.0.0 final is not published** — merging today would pin `main` to a prerelease. **New exit condition**: `vitest@5.0.0` final ships → bump the catalog off the prerelease → merge. The RC itself looks shippable (full verification below); this is a "don't pin main to a prerelease" hold, not a quality hold.

## What

Removes the `@vitest/browser` pnpm patch (patch file + `pnpm-workspace.yaml` declaration + lockfile regen) **and bumps vitest 4.1.10 → 5.0.0-rc.2**, which is the release that carries the patch's fix upstream. Per the original TODO, the patch is not removed against a version that still needs it.

The patch's `isEmbedded` channel guard protected against duplicate testers echoing `response:*` events. Two independent things now cover it: #422's spec-level prevention (no duplicate tester can open) and, as of 5.0.0-rc.1, the upstream fix itself.

## The upstream fix is genuinely in this release

vitest-dev/vitest#9381 merged upstream 2026-07-29 (commit `46a3cf7`) and ships first in **5.0.0-rc.1** (published 2026-08-11). No 4.1.x has it — the 4.1 line diverged before the merge and there was no backport.

Verified in the **installed tarball**, not just the release notes: the shipped `tester-Ck66x6M6.js` carries both guard points — the early `if (!isEmbedded) return` in the channel message listener, and the `if (isEmbedded)` gate on the `ready` post. The filename rotated from the patched `tester-BJtsJFpD.js`, which is exactly the `check:patches` trigger this PR predicted.

Re-verified after the rc.2 bump: the bundle rotated again (`tester-Ck66x6M6.js` → `tester-DDRabqAJ.js`) and still carries `isEmbedded = window.self !== window.top` with both guard points intact.

Note the shipped form is the reviewer-requested *remove-the-duplicate-participant* approach, **not** the `startsWith('response:')` symptom filter the local patch used — so the old regenerate-the-patch-on-bump recipe is dead, not merely unused.

## vitest 5 migration changes carried here

Three required, one cleanup:

1. **`test.browser.api` → `test.api`** — the browser server port is now a project-level option. This is a **hard type error**, so it surfaces via `typecheck`, not `test`: `navigation-location-plugin` tests passed while its typecheck failed. Three configs updated (`lit-ui-router`, `navigation-location-plugin`, `sample-app-shared`); the `VITEST_BROWSER_API_PORT` keying is preserved so concurrent `test` / `test:coverage` turbo tasks keep distinct dep-optimizer dirs.
2. **`@vitest/browser` is now an explicit devDep** in each browser-mode package. v5 puts `@vitest/browser/client` in the *client* `optimizeDeps.include`, but under pnpm's strict linking it was only reachable transitively via `@vitest/browser-playwright`, producing `Failed to resolve dependency` on every **cold** run. (The dep shape did not change between 4 and 5 — the optimizeDeps use did.) Warning-only, but it masks real failures, and it vanishes on a warm cache — so verify with `--force`.
3. **Inline projects now inherit the declaring config** unless they pass `extends: false` (only `name`, `projects`, and root `globalSetup` are excluded). The per-project `resolve.alias` duplication in `lit-ui-router/vitest.config.ts` — and the comment asserting inline projects do *not* inherit a root alias, now false under 5.x — is retired in favour of one root `resolve.alias`. `test:lit2-compat` is the teeth: `assertLitMajor` throws in both the happy-dom and browser projects if the swap silently fails, and it passes 169/169 with the alias declared only once. Per-project `cacheDir` overrides are deliberately **kept** — inheritance would otherwise collapse them onto a single dep-optimizer dir.

Audited and clean, no action needed: `clearMocks` now defaults `true` (the repo overrides it nowhere; 12 spec files use spies, all green, nothing relied on cross-test call accumulation). Coverage globs are unchanged in effect — v5's `coverageConfigDefaults` has no `include`/`all`, so the report is exactly what specs load; `lit-ui-router`'s four unreported files are barrels or types-only with zero runtime declarations, so 97.2% is complete rather than inflated.

## Why removal was safe even before the upstream fix (original evidence, retained)

- Pristine `@vitest/browser@4.1.10` **with** prevention: 3× `vitest run` (chrome + happy-dom) and 2× `--coverage`, plus turbo `test` and `test:coverage` — all 165/165, exit 0, **zero** `response:*` occurrences, no hangs.
- Control — pristine **without** prevention: immediate cascade, **134,894** `response:*` occurrences (`Error: Unknown event: response:response:ready` …), run failed. The prevention is the enabling factor, not coincidence.

With 5.0.0-rc.2 this is now belt *and* braces: prevention (#422/#429) plus the upstream guard.

## Verification on this branch

- `pnpm install` clean after the catalog + lockfile move; every vitest ref resolves to `5.0.0-rc.2` (0 remaining `rc.1` entries in the lockfile).
- `turbo run test typecheck --force` (zero cache, full repo) at rc.2: **74/75 tasks green**, **zero** deprecation warnings, **zero** `Failed to resolve` warnings, **zero** `response:*` echoes.
- The one non-green task, `sample-app-lit-e2e`, could not run locally: port 8787 was held by an unrelated `wrangler dev` from another worktree on this machine, so `start-server-and-test` got "server closed unexpectedly". Cypress e2e does not involve vitest, and it passed 5/5 at rc.1 — CI runs it isolated and is the arbiter here.
- `check:patches`: passes, "2 patches verified against installs" (`lit-dialog`, `@api-viewer/docs`) — `@vitest/browser` stays absent.
- `format:root` + `lint:package-json`: clean (32/32).
- After dropping the release-age exclusions (`7adb41b`): `pnpm install` clean on pnpm 11.21.0 with the lockfile byte-identical, `pnpm install --frozen-lockfile` clean, `check:patches` 2/2, `format:check:root` clean (35 files), `lint:package-json` 32/32.
- Earlier at rc.1: full `turbo run test typecheck --force` 75/75 including 5/5 Cypress e2e; `packages/lit-ui-router` test 169/169, coverage 169/169 (97.2% stmts), engines (firefox + safari) 50/50, lit2-compat 169/169.

## Release-age gate (scaffolding, now removed)

When the rc.2 bump landed the release was ~16h old, and pnpm's default `minimumReleaseAge` (1440min) plus this repo's `minimumReleaseAgeStrict: true` refused it outright — `ERR_PNPM_NO_MATURE_MATCHING_VERSION`, 35 versions across 9 packages, since bumping 4 catalog entries pulls in the whole `@vitest/*` family. That gate applies to CI's `--frozen-lockfile` install too, so the branch carried a temporary version-pinned `minimumReleaseAgeExclude` block (pinned rather than an `@vitest/*` glob, so a future fresh vitest release could not be silently waved through).

**The block is gone as of `7adb41b`.** The whole family published 2026-08-17T13:2xZ and is now >24h old, so the exclusions were inert; removing them restores the strict gate for every package. Verified resolution-neutral: `pnpm install` leaves `pnpm-lock.yaml` byte-identical (md5 `4bab08afa99eae4b7e99135ed4091413`) and `--frozen-lockfile` passes, so nothing was being held up by the block. Precedent: #258 dropped the TypeScript 7.0.2 exclusions the same way once they aged.

## Interaction with #421

#421's generic `check:patches` guard is patch-agnostic — after this lands it simply has one fewer resident (its inventory table's `@vitest/browser` row becomes historical). The guard did its job here exactly as designed: the bump rotated the hashed filename and forced the look.

## TODO

- [x] Bump `@vitest/browser` to a release that includes the upstream fix the patch carries (vitest-dev/vitest#9381), then drop the patch — done via `5.0.0-rc.1`; the patch was never removed against a version that still needed it.
- [x] Delete the `minimumReleaseAgeExclude` block once rc.2 ages past the 24h window — done in `7adb41b`; verified lockfile-neutral and `--frozen-lockfile` clean.
- [ ] Wait for `vitest@5.0.0` final, bump the catalog off the prerelease, then unpark and merge.


packages/navigation-location-plugin/vitest.config.ts
```diff
@@ -60,12 +60,13 @@ export default defineConfig({
           name: 'browser',
           globals: true,
           include: browserOnlySpecs,
+          // vitest 5 moved the browser server port to test.api
```
simshanith:
```suggestion
```

packages/lit-ui-router/vitest.config.ts
```diff
@@ -55,18 +55,18 @@ export default defineConfig({
       },
       {
         cacheDir: `node_modules/.vite-${cacheKey}-browser`,
-        resolve: litResolve,
         test: {
           name: 'browser',
           globals: true,
           setupFiles: ['./vitest.setup.ts', './vitest.setup.browser.ts'],
           include: browserOnlySpecs,
+          // vitest 5 moved the browser server port to test.api
```
simshanith:
```suggestion
```

packages/lit-ui-router/vitest.config.ts
```diff
@@ -12,22 +12,23 @@ const browserOnlySpecs = ['src/specs/ui-sref.spec.ts'];
...
-// root and subpath — to the lit-2 alias devDep. Applied per project: inline
-// projects do NOT inherit a root-level resolve.alias. The VITE_ prefix
-// carries the same variable into import.meta.env everywhere (browser
-// included), where vitest.setup.ts asserts the swap took.
+// root and subpath — to the lit-2 alias devDep. Declared once at the root:
+// since vitest 5 inline projects inherit the declaring config unless they
+// opt out with `extends: false`. The VITE_ prefix carries the same variable
+// into import.meta.env everywhere (browser included), where vitest.setup.ts
+// asserts the swap took — so a lit2-compat run is the teeth on that
+// inheritance, not an assumption about it.
```
simshanith:
```suggestion
// asserts the swap
```

apps/sample-app-shared/vitest.config.ts
```diff
@@ -8,12 +8,13 @@ export default defineConfig({
     setupFiles: ['./vitest.setup.ts'],
     // hanging-process logs the open handles in CI
     reporters: process.env.CI ? ['default', 'hanging-process'] : ['default'],
+    // vitest 5 moved the browser server port to test.api
```
simshanith:
```suggestion
```

PR #441, opened 2026-07-23
chore(mise): move npm audit/fund opt-out to setup task env

simshanith:
`examples/package.json` repeated `--no-audit --no-fund` on each of the three `npm ci --prefix` legs. Those run from pnpm's `postinstall`, which only ever runs under `mise run setup` (every CI workflow and the CONTRIBUTING flow), so the setup task's env reaches them.

```toml
[tasks.setup.env]
CYPRESS_INSTALL_BINARY = "0"
NPM_CONFIG_AUDIT = "false"
NPM_CONFIG_FUND = "false"
```

Verified locally, both directions:

- `mise run setup` → `added 27 packages in 2s` (no audit/funding lines) with the flags removed.
- Same `npm ci` with `env -u NPM_CONFIG_AUDIT -u NPM_CONFIG_FUND` → `added 27 packages, and audited 28 packages` + `8 packages are looking for funding`.

Note: values are TOML strings (`"false"`), matching the existing `CYPRESS_INSTALL_BINARY = "0"`; npm parses the env value itself.

### DIFF
--- a/.config/mise/config.toml
+++ b/.config/mise/config.toml
@@ -65,6 +65,10 @@ run = 'pnpm install ${usage_release:+--filter @tools/release...}'
 
 [tasks.setup.env]
 CYPRESS_INSTALL_BINARY = "0"
+# npm config for the examples/* `npm ci --prefix` postinstall legs; env beats
+# per-call flags, so the example manifests stay flag-free.
+NPM_CONFIG_AUDIT = "false"
+NPM_CONFIG_FUND = "false"
 
 [tasks.lint_actionlint]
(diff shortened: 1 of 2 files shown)


.config/mise/config.toml
```diff
@@ -65,6 +65,10 @@ run = 'pnpm install ${usage_release:+--filter @tools/release...}'
 
 [tasks.setup.env]
 CYPRESS_INSTALL_BINARY = "0"
+# npm config for the examples/* `npm ci --prefix` postinstall legs; env beats
+# per-call flags, so the example manifests stay flag-free.
```
```suggestion
# npm config for the examples/*
```

PR #453, opened 2026-07-24
build(mise): add taplo as the TOML formatter with a turbo format gate

simshanith:
## Why

`oxfmt` formats every filetype in the repo **except** `.toml` (its globs are `md,ts,js,json,yaml,...`). So `.config/mise/config.toml` and the member `tools/*/mise.toml` files had no format gate and could drift — which is how #452's hand-edited TOML kept needing a manual `taplo fmt --check`. This adds the missing tool + CI gate.

## What

- **Tool pin.** `aqua:tamasfe/taplo = 0.10.0` in `[tools]`, alongside the other aqua tools (actionlint/zizmor), locked across all 7 platforms.
- **mise tasks.** `format_toml` / `format_check_toml` run taplo; package.json `format:toml` / `format:check:toml` delegate via `mise run`, exactly like `lint:workflows`.
- **turbo graph.** `//#format:toml` / `//#format:check:toml` join the existing `format` / `format:check` `with`-groups, so they run in parallel with the oxfmt root task and flow into `ci:pull_request` automatically — no workflow edit needed (CI installs aqua tools via mise, same as actionlint/zizmor).

## Two gotchas the gate surfaced (both verified)

**taplo does not skip node_modules.** Pre-install it found 3 files; post-`pnpm install` it found **207**, pulling in workspace-symlinked `mise.toml` and `release-it` test stubs under `node_modules`. Scoped via a `.config/taplo.toml` (pointed to by mise `[env]` `TAPLO_CONFIG`, since `.config/` is not on taplo auto-discovery — this also scopes ad-hoc `taplo fmt`) with a top-level `exclude = ["**/node_modules/**"]` (default include = all .toml, minus the vendored tree — new repo .toml anywhere is picked up automatically). Note: `include`/`exclude` are **top-level** keys, not nested under `[files]` (a first attempt failed that way). Verified a deliberately-malformed `node_modules/*.toml` is excluded (exit 0) while the 4 repo files are checked.

**taplo has no checksums file and no attestations.** So `mise lock` could only checksum macos-arm64 (the platform it downloads locally, blake3); the other 6 locked url-only, leaving linux CI unverified — the exact trap from prior `mise.lock` platform-stripping incidents. Computed sha256 for all 7 platforms from the release assets and injected them; verified mise re-verifies against the sha256 on a forced reinstall (it hashes the same `.gz`).

## Verification

- `turbo run format:check` → exit 0, taplo scans exactly 3 files
- `turbo run lint` → exit 0 (a stale-worktree `check-patches` failure turned out to be missing deps, fixed by `mise run setup`, unrelated to this change)
- `turbo run format:check --force` → `found files total=3`, no node_modules
- forced `mise uninstall && install` → checksum step verifies against the locked sha256

## Relationship to #452

Separate PR by request — #452 (remote-cache signing) stays scoped; this is toolchain + CI-graph infra with its own blast radius (tool lockfile). #452 keeps its manual `taplo fmt --check` until this merges.

## Update: taplo lint in both pipelines + node_modules-walk perf fix

- **`taplo lint` added to `lint_workflows`** (3rd leg beside actionlint/zizmor), so taplo now runs in **both** `mise run lint_workflows` and the **turbo ci** graph (via the existing `//#lint:workflows` node — confirmed in `turbo run ci --dry`). Redundant with the `fmt --check` format gate by design (the two gates independently re-walk the config tree and backstop each other's wiring). `taplo lint` runs **offline / syntax-only** by default (no schema catalogs fetched); schema-aware key/type validation is left opt-in (needs a vendored schema to stay hermetic).
- **perf: taplo config switched from broad-include + `exclude` to shallow per-root `include` globs.** taplo has no gitignore/vcs pruning, so a broad `**/*.toml` walked the entire `node_modules` tree (~14s for 4 real files) on every fmt/lint run, local and CI. Anchored shallow globs let taplo prune `node_modules` → **~0.07s (~200×)**, full coverage of every workspace member root. Deeply nested TOML would need its own glob (documented in `.config/taplo.toml`).

## Update 2: implement file selection via `git ls-files` (supersedes the shallow-include + coverage-check above)

Rather than maintain shallow include globs *and* a check to guard them, the tasks now pass taplo an **explicit file list** from `git ls-files -- '*.toml'`. That set *is* "`**/*.toml` minus node_modules" (node_modules is untracked), produced with no directory walk and no `pnpm install` — so it's **~0.2s, complete by construction, zero glob maintenance, no reconciliation**, install-independent.

- Dropped the per-root include globs and **removed `//#check:toml-coverage`** (and its `ci:pull_request` node / root script) — no heuristic left to check.
- `.config/taplo.toml` keeps only `exclude = ["**/node_modules/**"]` as a safety net for an ad-hoc bare `taplo` (no file args → it would otherwise reformat ~200 vendored stubs).
- Task body: `cd "$(git rev-parse --show-toplevel)"; files=$(git ls-files -- '*.toml'); [ -z "$files" ] || taplo fmt $files` — applies to `format_toml`, `format_check_toml`, and `lint_toml`.

## Update 3: shared `taplo` mise task + review suggestion

- Extracted the file-selection logic into one shared task `.config/mise/tasks/taplo`. It mirrors taplo's CLI: `<cmd>` (fmt | lint | check) passes through **verbatim** — note `taplo check` is schema validation, *not* `fmt --check` — with `--check` a fmt-only flag, and trailing files overriding the `git ls-files` default (so `mise run taplo fmt --check -- path/to/x.toml` targets a single file for editor/pre-commit use). `format_toml`/`format_check_toml`/`lint_toml` are now thin aliases.
- No `cd` in the task — mise runs tasks from config_root regardless of invocation cwd (verified).
- Applied the review suggestion shortening the relocated-`taplo.toml` comment with the taplo config-docs link.


.config/mise/config.toml
```diff
@@ -18,6 +18,11 @@ config_roots = ["tools/*"]
 # aren't at root), matching pnpm-script resolution; elsewhere it's a dead
 # entry, and at the repo root a harmless duplicate of the config_root one.
 _.path = ["{{cwd}}/node_modules/.bin", "{{config_root}}/node_modules/.bin"]
+# taplo's config lives under .config/ (not the auto-discovered root .taplo.toml),
+# so point every taplo run at it here. The tasks scope files via `git ls-files`;
+# this only gives an ad-hoc bare `taplo` its node_modules exclude. Absolute so any
+# cwd resolves it.
```
```suggestion
# Relocated taplo.toml out of root into `.config/`
# https://taplo.tamasfe.dev/cli/usage/configuration.html#configuration-file
```

.config/mise/tasks/taplo
```diff
@@ -0,0 +1,30 @@
...
+#USAGE arg "[file]…" help="TOML files to target; default: every tracked *.toml"
+#
+# Shared taplo entrypoint. <cmd> passes through to taplo verbatim — note `taplo
+# check` is schema validation, NOT `fmt --check` — with --check a fmt-only flag.
+# Default files = `git ls-files -- '*.toml'`: the broad **/*.toml-minus-node_modules
+# set with no directory walk and no pnpm install (taplo has no gitignore pruning,
+# so a bare walk is ~14s and hits node_modules). Given files win, and explicit
+# paths bypass taplo's walker too, so a single-file editor/pre-commit run is
+# `mise run taplo fmt --check -- path/to/x.toml`. mise runs tasks from config_root,
+# so no cd is needed.
```
simshanith:
```suggestion
```

.config/mise/tasks/taplo
```diff
@@ -0,0 +1,30 @@
...
+
+set -- "$usage_cmd"
+[ "${usage_check:-}" = "true" ] && set -- "$@" --check
+
+files="${usage_file:-}"
+[ -n "$files" ] || files=$(git ls-files -- '*.toml')
+[ -n "$files" ] || exit 0
+
+# unquoted $files: split the space/newline-joined list into args; TOML paths have no spaces
+exec taplo "$@" $files
```
simshanith, line 23: exec needs absolute path no?

.config/mise/tasks/taplo
```diff
@@ -0,0 +1,30 @@
...
+  fmt | lint | check) ;;
+  *) echo "taplo: unknown cmd '$usage_cmd' (fmt | lint | check)" >&2; exit 2 ;;
+esac
+
+set -- "$usage_cmd"
+[ "${usage_check:-}" = "true" ] && set -- "$@" --check
+
+files="${usage_file:-}"
+[ -n "$files" ] || files=$(git ls-files -- '*.toml')
+[ -n "$files" ] || exit 0
```
simshanith: what is taplo's exit for this condition? seems like it should be safe to call; print warning to stderr instead maybe

.config/mise/tasks/taplo
simshanith: No — verified. `exec taplo` does a normal PATH lookup (mise puts the aqua shim on PATH), so no absolute path is needed for the binary. The file args are relative paths from `git ls-files`, and they resolve against the task cwd, which mise sets to config_root regardless of where `mise run` was invoked — I ran `//:taplo fmt --check` from `docs/` and it still collected the 4 tracked TOMLs correctly (cwd=config_root in the log). Left as-is.

.config/mise/tasks/taplo
simshanith: Done in 60ae95d. taplo with no file args falls back to its config walk (the ~14s node_modules walk we are avoiding), so we do not want to just call it bare. Kept the `exit 0` guard but it now prints `taplo: no tracked *.toml files matched; nothing to do` to stderr first, so the can't-happen case surfaces instead of passing silently.

.config/mise/tasks/taplo
simshanith: Applied in 60ae95d — dropped "so no cd is needed" and ended the sentence at config_root (kept that fact since it is what makes the relative `git ls-files` paths resolve, per the exec thread above).

.config/mise/config.toml
simshanith: Already applied — the comment now reads exactly this (relocated-out-of-root note + config-docs link).

.config/mise/tasks/taplo
simshanith: Correction — took the whole 6–14 block out (a75e487), not just the tail. You are right it restated the USAGE directives and was doc-rot prone. The one fact that earns its keep (bare taplo → ~14s node_modules walk) survives as the inline comment on the empty-files guard, where it explains the code.

PR #454, opened 2026-07-24
chore(lit-ui-router): ship @oxc-project/runtime as >=0.50.0

simshanith:
## What

`@oxc-project/runtime` is a genuine **runtime** dependency — the oxc analog of `tslib`. The shipped `dist/ui-router.js` and `dist/ui-view.js` do `import _decorate from "@oxc-project/runtime/helpers/decorate"`, emitted by `oxc-transform` (default `HelperMode.Runtime`) for the legacy Lit decorators. It belongs in `dependencies`, but the exact `0.140.0` pin over-constrained consumers (duplicate installs, no patch dedup).

This widens the **shipped** range to the empirically guaranteeable floor, while keeping an exact pin around purely as a dependabot signal.

## The floor — how `>=0.50.0` was derived

A **throwaway** bisection over all 115 published versions (`0.49.0` → `0.141.0`), installing each and resolving `helpers/decorate` under nodenext ESM:

| version | result |
|---|---|
| `0.49.0` | ❌ **broken publish** — exports map refs `helpers/decorate.js` / `helpers/esm/decorate.js` but ships neither file (`ERR_MODULE_NOT_FOUND`) |
| `0.50.0` … `0.141.0` | ✅ resolve a **stable arity-4** `decorate` function throughout |

The only helper we emit is `helpers/decorate` (one call site each in the two decorated custom elements), so the contract surface is tiny and the floor is `>=0.50.0`, **unbounded above by design**. The bisection matrix itself is not kept — its evidence lives here + in CI history.

## How it ships (no publish-side code)

- New **`publishedDependencies`** catalog holds `@oxc-project/runtime: '>=0.50.0'` — the consumer-facing-range analog of the existing `publishedPeer` catalog.
- `pnpm pack` resolves the `catalog:` ref **verbatim** (exactly as `lit: catalog:publishedPeer` ships `^3.0.0`), so no change to `strippedManifest` / the release pack.

## The retained exact pin (signal only)

Because dependabot on pnpm 11 is unreliable at regenerating lockfiles (version-update PRs get superseded), the pin isn't a bump mechanism — it's kept for **dependabot version/security visibility** on a concrete version:

- default catalog: `oxc-project-runtime: 'npm:@oxc-project/runtime@0.140.0'` (aliased)
- referenced as a `devDependency` (`"oxc-project-runtime": "catalog:"`)
- `strippedManifest()` removes all devDependencies at publish, so the alias **never reaches consumers**
- pnpm **dedupes** the wide `>=0.50.0` dependency onto that pinned `0.140.0`, keeping dev/CI anchored to the `oxc-transform`-aligned version. The real forward signal is the normal build + decorator tests re-running whenever oxc is bumped manually.

## Verification

- `pnpm pack` → published `dependencies` = `{"@oxc-project/runtime":">=0.50.0","@uirouter/core":"^6.0.8","lit":"^3.0.0"}`
- `strippedManifest()` → devDependencies gone; alias does **not** leak to consumers
- lint (incl. `eslint-plugin-pnpm` catalog rules), typecheck, and the **140 happy-dom tests** pass
- ships-affecting manifest change → expect `check:published-diff` to flag it as a release (intentional)

## Deferred (follow-up, easily added)

A slim, dependabot-independent **floor guard** (`tools/runtime-backtest`, one `@0.50.0` alias) that dist-greps the emitted `helpers/*` and asserts each resolves against the floor — the runtime analog of `dts-backtest`. Left out to keep this PR to the range change; the `>=0.50.0` promise is currently defended by the bisection evidence + the fact that a floor-raise requires a deliberate build-config change (target/decorator mode).


pnpm-workspace.yaml
```diff
@@ -10,7 +10,10 @@ catalog:
   '@api-viewer/docs': 1.0.0-pre.10
   '@arethetypeswrong/core': ^0.18.5
   '@codecov/bundle-analyzer': ^2.0.1
-  '@oxc-project/runtime': 0.140.0
+  # exact pin, kept only for dependabot version/security signal; referenced as a
+  # devDep alias (stripped at publish). lit-ui-router SHIPS the wide
+  # catalog:publishedDependencies range instead.
```
```suggestion
 ```

File: pnpm-workspace.yaml [hunk 2]
```diff
@@ -86,6 +89,12 @@ catalogs:
   # fails the typecheck:peer-floor task when the two drift apart
   peerFloor:
     lit-ui-router-floor: npm:lit-ui-router@1.7.0
+  # consumer-facing ranges for shipped hard (non-peer) dependencies. Widened as
+  # far down as the emitted oxc helpers resolve: 0.49.0 is a broken publish
+  # (exports map refs helpers/decorate but ships no file), 0.50.0+ resolve a
+  # stable arity-4 decorate through 0.141.0. Unbounded above by design.
```
```suggestion
```

PR #456, opened 2026-07-24
build(shellcheck): gate the repo shell surface via lint_workflows

simshanith:
## What

Adds a **shellcheck gate** for the repo's shell scripts. No shellcheck existed anywhere before this (no config, no task, no CI leg). Wires it into the existing `lint_workflows` job (rides `lint-workflows.yml` alongside actionlint, zizmor, and taplo) — **no new workflow file**.

Rebased onto the merged taplo TOML gate (#453) and mirrors its shape:
- `aqua:koalaman/shellcheck = 0.11.0` pin + all-7-platform `mise.lock` checksums (aqua ships checksums, so no manual restore; linux-x64 verified against the real asset).
- Shared file task `.config/mise/tasks/shellcheck` owns file selection; a thin `config.toml` alias (`lint_shellcheck`) keeps the turbo/task name stable and joins the `lint_workflows` depends list.
- `//#lint:workflows` turbo inputs extended to also track `**/*.sh`, the extensionless mise task scripts (`.config/mise/tasks/**`), and the tool pin.

## File selection

**Shebang-based** (`git ls-files` first-line scan over `*.sh`/`*.bash` + extensionless files), not extension-only — so it covers both `*.sh` and the repo's extensionless mise-task convention. Candidates are narrowed to `*.sh`/`*.bash`/extensionless *before* the shebang read, so minified `.js` bundles and binaries are never opened. git ls-files already omits node_modules.

The two **vendored, verbatim-from-upstream** scripts under `apps/sample-app-shared/public/static/data/` are **excluded** from the default set — third-party static-asset scripts, not repo tooling, and they carry an upstream style nit (SC2129) we don't own. Still checkable on demand: `mise run shellcheck -- <path>`. Empty set warns and passes (main today ships only the excluded pair, so the gate is green and activates when PRs add scripts).

## Refactor, don't disable

The gate discovers the **mise task scripts themselves**, so both `.config/mise/tasks/shellcheck` and `.config/mise/tasks/taplo` are held to it. Each builds a tracked-file list and passes it to its tool — the classic unquoted-`$files` word-split (SC2086/SC2206). Rather than suppress with inline `# shellcheck disable=`, both tasks are refactored to build a **bash array** (`read -ra` for the given-files arg, `mapfile` for discovery) and pass `"${files[@]}"` — quoted, split-safe, no directives. (The taplo task change is the only edit to merged code; it's behavior-preserving — verified below.)

No repo-wide `.shellcheckrc`: the gate runs at default severity with no global suppressions.

## Verification (local, on rebased branch)

- ✅ `mise run lint_workflows` — actionlint + zizmor + taplo + shellcheck all green.
- ✅ shellcheck discovers + lints both mise task scripts clean (extensionless coverage works).
- ✅ taplo still processes its 4-file TOML set and `fmt --check` passes; given-files mode works (array refactor is behavior-preserving under `set -euo pipefail`).
- ✅ Negative test: a probe script with `[ $foo == "x" ]` fails the gate (SC2154/SC2086).
- ✅ `taplo fmt --check` clean on the edited `config.toml`.
- ✅ linux-x64 lock checksum matches the published asset.


.config/mise/tasks/shellcheck
```diff
@@ -0,0 +1,36 @@
...
+    IFS= read -r line <"$f" || continue
+    case "$line" in
+    '#!'*/sh | '#!'*/sh\ * | '#!'*/bash | '#!'*/bash\ * | '#!'*env\ sh* | '#!'*env\ bash*)
+      files+=("$f")
+      ;;
+    esac
+  done < <(git ls-files -- '*' ":!:$VENDORED/*")
+fi
+
+# Tracked set can be empty (main ships only the excluded vendored pair); warn, don't fail.
```
simshanith:
```suggestion
# warn on empty set
```

.config/mise/config.toml
```diff
@@ -30,6 +30,9 @@ idiomatic_version_file_enable_tools = ["node"]
 "aqua:rhysd/actionlint" = "1.7.12"
 # GitHub Actions security auditor; complements actionlint's syntax checks
 "aqua:zizmorcore/zizmor" = "1.26.1"
+# Shell-script static analysis; owns the repo's shell surface (mise task
+# scripts, *.sh). Pinned like the other aqua tools.
```
```suggestion
# Shell-script static analysis
```

.config/mise/config.toml
```diff
@@ -93,11 +96,17 @@ run = "zizmor --offline .github/workflows"
 description = "Validate tracked TOML structure (mise configs, member mise.toml) with taplo"
 run = "mise run taplo lint"
 
+[tasks.lint_shellcheck]
+# Thin alias; file selection + rationale live in the `shellcheck` file task
+# (.config/mise/tasks/shellcheck), keeping the turbo/lint_workflows names stable.
```
```suggestion
# Thin alias; file selection + rationale live in the `shellcheck` file task
# (.config/mise/tasks/shellcheck)
```

PR #462, opened 2026-07-25
ci(build-test): opt into the main graph by branch prefix

simshanith:
Stacked on #458 (base is its branch, not `main`).

## Why

The main-only guards — `test:engines`, `@tools/release#check:pack`, the full `dts-backtest` TypeScript matrix — run in `ci:main`, which only fires on pushes to `main`. So a break in them surfaces *after* merge, on main, rather than on the PR that caused it. This adds a way to pull that signal forward.

## Two paths

**Per run — already worked, now documented.** `workflow_dispatch` takes a ref, so **Build and Test** with `mainGraph: true` has always been able to run the main graph against an arbitrary branch. Nothing needed to change; it just wasn't written down. Verified live before touching anything (see below).

**Per branch — new.** A branch named `ci-main/<topic>` builds `ci:main` on every push. It runs *even when the branch merges cleanly*: the `pull_request` run that covers that SHA covers only the PR graph, which is precisely what the opt-in is asking past. The prefix is the entire opt-in — no input, no label, no commit trailer.

Tagging and publishing stay bound to pushes to `main`, so neither path can release.

## The refactor this required

`build-test-run.yml` was inferring the graph from `github.event_name`/`github.ref`. With a third caller that inference is no longer possible — and it was never really its question to answer, since each caller knows its own event. The graph choice moves to the callers and the shared body now keys purely on `inputs.mainGraph`:

| caller | `mainGraph` |
| --- | --- |
| `build-test.yml` (PR) | `false` |
| `build-test.yml` (push to `main`) | `true` |
| `build-test.yml` (dispatch) | the input |
| `build-test-branch.yml` | the gate's output |

**This fixes a latent hole rather than only enabling the feature.** The "Install Playwright OS deps (firefox+webkit)" step keyed on that same event predicate. Any non-main caller running `ci:main` would have run `test:engines` *without* its OS deps — so the branch opt-in would have been broken on a cache hit if the predicate had been left alone. Same for the Cypress-video upload, which now keys on its input alone.

## Where the prefix lives

One place: `decide()` in `branch-ci-gate.core.ts`, checked *before* mergeability since it is the one case where a covered SHA still needs its own run. Consequences:

- `build-test-branch.yml` needs no extra `if` term — the gate already answers `run=true` for the opt-in, so the build job's existing condition covers it.
- The fail-open path re-derives it from `GITHUB_REF_NAME`. A `startsWith` cannot be what broke the gate, and degrading it would silently run the smaller graph — the opposite of what was asked for.
- `ci-main` and `ci-maintenance` do **not** opt in; the separator is required.

## Verification

Live, against real runs:

| case | evidence |
| --- | --- |
| dispatch runs the main graph on an arbitrary branch | run [30169092107](https://github.com/simshanith/lit-ui-router/actions/runs/30169092107) — ref `throwaway/no-pr-trigger-split`, `success`, executed `mise run ci_main`. This is the *pre-change* behaviour, confirmed before editing. |
| `ci-main/` branch push builds `ci:main` | `ci-main/verify-optin` @ `7e6d70f`, run [30169324417](https://github.com/simshanith/lit-ui-router/actions/runs/30169324417) — `success`. Gate logged `RUN: `ci-main/` branch prefix — running the ci:main superset`; the build job executed `mise run ci_main` and ran `test:engines` + `check:pack`, i.e. the main-only guards, on a branch push. |
| trigger isolation holds | only **Build and Test (branch)** fired on that push — two rows, `build_and_test (signal gate)` and `build_and_test (branch) / run`. |

Unit-tested (`node --test`, 30 cases in the gate suite, 0 fail):

- opt-in wins over a clean-mergeable verdict — the case a live no-PR branch cannot demonstrate, since "no open PR" would run it anyway
- the opt-in graph survives every other run reason (conflict, indeterminate, no probe, no PR)
- prefix anchoring and the separator requirement

Also green locally: `tsc --noEmit`, `oxlint`, `oxfmt --check`, `actionlint`, `zizmor`, `taplo`.

## Docs

`TURBO.md` gains a "Smoke-testing the main graph before merge" section; `RELEASE.md` notes both paths where it describes the guards that gate a release.

## Merge notes

- Merge #458 first — this is stacked on it and inherits its ruleset repoint (`build_and_test` → `build_and_test / run`).
- Throwaway branches to delete after merge: `ci-main/verify-optin`, `throwaway/no-pr-trigger-split` (the latter is #458's evidence ref).

### DIFF
--- a/tools/build_and_test/src/branch-ci-gate.core.ts
+++ b/tools/build_and_test/src/branch-ci-gate.core.ts
@@ -22,7 +22,19 @@ export type BaseVerdict = {
   detail?: string;
 };
 
-export type Decision = { run: boolean; reason: string };
+export type Decision = { run: boolean; reason: string; mainGraph: boolean };
+
+/**
+ * Branch-name prefix that opts a push into the `ci:main` superset — the
+ * main-only guards a pull_request run never covers, smoke-tested before merge
+ * rather than after. Named for the task it selects (`mise run ci_main`).
+ */
+export const MAIN_GRAPH_PREFIX = 'ci-main/';
+
+/** Whether this branch name opts into the main graph. */
+export function wantsMainGraph(branch: string): boolean {
+  return branch.startsWith(MAIN_GRAPH_PREFIX);
+}
 
 /**
  * Validate `gh pr list --json number,baseRefName` output. Throws on any shape
@@ -95,10 +107,22 @@ function listBases(verdicts: readonly BaseVerdict[]): string {
 export function decide(
   prs: readonly OpenPr[],
   verdicts: readonly BaseVerdict[],
+  mainGraph = false,
 ): Decision {
+  // Checked before mergeability, because this is the one case where a branch a
+  // pull_request run *does* cover still needs its own: that run builds the PR
+  // graph, and the main-only guards are exactly what the opt-in asked for.
+  if (mainGraph) {
+    return {
+      run: true,
+      mainGraph,
+      reason: `\`${MAIN_GRAPH_PREFIX}\` branch prefix — running ci:main`,
+    };
+  }
   if (prs.length === 0) {
     return {
       run: true,
+      mainGraph,
       reason:
         'no open PR has this branch as its head — no pull_request run will cover this push',
     };
@@ -106,6 +130,7 @@ export function decide(
   if (verdicts.length === 0) {
     return {
       run: true,
+      mainGraph,
       reason: `${prs.length} open PR(s) but no merge probe ran — running rather than assume coverage`,
     };
   }
@@ -115,6 +140,7 @@ export function decide(
   if (conflicting.length > 0) {
     return {
       run: true,
+      mainGraph,
       reason: `conflicts with ${listBases(conflicting)} — GitHub builds no merge ref, so the pull_request run is skipped`,
     };
   }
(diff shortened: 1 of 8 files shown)


tools/build_and_test/src/branch-ci-gate.core.ts
```diff
@@ -95,17 +107,30 @@ function listBases(verdicts: readonly BaseVerdict[]): string {
...
+  mainGraph = false,
 ): Decision {
+  // Checked before mergeability, because this is the one case where a branch a
+  // pull_request run *does* cover still needs its own: that run builds the PR
+  // graph, and the main-only guards are exactly what the opt-in asked for.
+  if (mainGraph) {
+    return {
+      run: true,
+      mainGraph,
+      reason: `\`${MAIN_GRAPH_PREFIX}\` branch prefix — running the ci:main superset, which no pull_request run covers`,
```
simshanith:
```suggestion
      reason: `\`${MAIN_GRAPH_PREFIX}\` branch prefix — running ci:main`,
```

simshanith: Applied — reason is now ```ci-main/`` branch prefix — running ci:main`. Test assertion follows it (`/running ci:main/`).

The `report`/`summaryMarkdown` arity fix went to the stack base instead: #458 now carries `refactor(build_and_test): pass the gate run as one value` (86c69e8), since the five positionals originated there. This branch merged it back in, so its own diff is opt-in only.

PR #526, opened 2026-08-08
perf(turbo): parallelize docs:api by moving ordering onto the docs package

simshanith:
## What

The root `docs:api` task carried a `^docs:api` self-chain, which serialized typedoc across the workspace dependency graph. On a cold `turbo run typecheck --force` roughly 95% of wall time was spent waiting on typedoc, with `lit-ui-router-mobx#docs:api` unable to start until `lit-ui-router#docs:api` finished — even though typedoc reads sources and needs only its dependencies' *builds*.

This drops the self-chain and moves the ordering requirement onto the one package that actually has it.

## Why the self-chain was NOT boilerplate — read before re-deleting it

The obvious move is to delete `^docs:api` from the root task and stop. **That breaks the build**, and it breaks it in a way a warm working tree hides (the `docs/api/*` output is already on disk, so the build passes anyway).

`docs`'s only *direct* workspace dependency among the libraries is `ui-router-server`. `lit-ui-router`, `lit-ui-router-mobx` and `ui-router-navigation-location-plugin` reach `docs` only **transitively**, via the sample apps. Turbo's `^task` reaches direct dependencies only, so the `^docs:api` self-chain was the mechanism propagating the ordering transitively — exactly like `build:types`, which carries the comment *"self-chain to build types transitively"*.

Delete it alone and `docs#build` fails with:

```
[UNRESOLVED_IMPORT] Could not resolve '../api/lit-ui-router-mobx/typedoc-sidebar.json' in .vitepress/config.ts
```

So the requirement is moved, not removed: `docs/.vitepress/config.ts` statically imports four `typedoc-sidebar.json` files, and `docs`'s tasks now name those four producers with package-qualified edges.

## The five affected docs tasks

| task | before | after |
| --- | --- | --- |
| `build` | `^docs:api` | 4 explicit `<pkg>#docs:api` edges |
| `typecheck` | `^docs:api` | 4 explicit `<pkg>#docs:api` edges |
| `docs` (dev server) | `^docs:api` | 4 explicit `<pkg>#docs:api` edges |
| `docs:preview` | `^docs:api`, `build` | `build` — inherits the edges through `build` |
| `wrangler:dev` | `^docs:api`, `build` | `build` — inherits the edges through `build` |

No package other than `docs` depended on `^docs:api` (verified by grepping every `turbo.json` in the repo). The root `docs` task keeps its `^docs:api`, but `docs` is the only package with a `docs` script and it overrides the root entry outright, so that line is inert either way.

## Measurements

`--force`, warm install, 3 runs each:

| command | before | after |
| --- | --- | --- |
| `turbo run docs:api --force` | 7.83s / 8.23s / 8.56s | **6.41s / 6.04s / 6.31s** |
| `turbo run typecheck --force` | 8.89s / 8.23s / 7.09s | **6.07s / 5.81s / 5.37s** |

All runs exit 0 with `Tasks: N successful, N total` and no `Failed:` line. `typecheck` also drops from 45 to 42 executed tasks: `build:js` for mobx / nav-plugin / server were being dragged in purely as a side effect of the phantom `docs:api` nodes' `^build`, and typechecking never needed them.

### Overlap proof (`--summarize`, offsets from first task start)

Before:

```
lit-ui-router#docs:api                        start=598ms   end=3689ms
lit-ui-router-mobx#docs:api                   start=3689ms  end=5430ms   <- starts exactly when lit-ui-router ends
ui-router-navigation-location-plugin#docs:api start=598ms   end=2974ms
ui-router-server#docs:api                     start=598ms   end=2177ms
```

After:

```
lit-ui-router#docs:api                        start=628ms   end=4719ms
lit-ui-router-mobx#docs:api                   start=1955ms  end=4830ms   <- starts when lit-ui-router#build lands
ui-router-navigation-location-plugin#docs:api start=628ms   end=4499ms
ui-router-server#docs:api                     start=628ms   end=3181ms
```

Individual task durations go *up* because four typedoc processes now contend for CPU; wall time goes down.

## Byte-identical output

```
find docs/api -type f | sort | xargs shasum
```

107 files before, 107 files after, `diff` empty — including a run performed after deleting all four generated `docs/api/*` subdirectories, so the ordering was exercised cold rather than masked by stale output.

## Ordering proven by construction, not by a lucky green run

`turbo run docs#typecheck --dry=json` lists all four producers in `docs#typecheck`'s dependency set:

```
docs#typecheck <- [..., lit-ui-router#docs:api, lit-ui-router-mobx#docs:api,
                   ui-router-navigation-location-plugin#docs:api, ui-router-server#docs:api]
```

The same four appear in the `--dry=json` graphs for `docs#build`, `docs#docs`, `docs#docs:preview` and `docs#wrangler:dev`, and in `turbo run ci:pull_request --dry=json` (318 tasks, and exactly those four `#docs:api` nodes — the 13 no-op ones the self-chain used to synthesize are gone).

## Verification

- `turbo run docs:api --force` — green, output byte-identical
- `turbo run docs#build --force` with `docs/api/*` deleted — green (the cold path that would have caught a broken graph)
- `turbo run build --force` — green
- `turbo run typecheck --force` — green
- `turbo run build typecheck lint format:check` — 106/106 green
- `turbo run ci:pull_request --dry=json` — graph resolves, four `#docs:api` nodes present

Not run locally: `e2e` / cypress, and `docs:preview` / `wrangler:dev` execution (config inspection + `--dry=json` only).


docs/turbo.json
```diff
@@ -21,26 +21,40 @@
       "dependsOn": ["bundle:worker"]
     },
     "docs:preview": {
-      "dependsOn": ["^docs:api", "build"],
+      // typedoc sidebars ride in via build
```
simshanith:
```suggestion
```

docs/turbo.json
```diff
@@ -21,26 +21,40 @@
...
+      "dependsOn": ["build"],
       "cache": false,
       "persistent": true
     },
     "wrangler:dev": {
-      "dependsOn": ["^docs:api", "build"],
+      // typedoc sidebars ride in via build
```
simshanith:
```suggestion
```

docs/turbo.json
```diff
@@ -21,26 +21,40 @@
...
+      "dependsOn": ["build"],
       "cache": false,
       "persistent": true
     },
     "build": {
+      // .vitepress/config.ts statically imports each typedoc-sidebar.json;
+      // only ui-router-server is a direct dep, so `^docs:api` can't reach the
+      // rest — name every producer explicitly.
```
simshanith: wondering if we should make them dev deps here instead so ^docs:api does reach them

PR #527, opened 2026-08-08
test(docs): assert every docs:api producer is a direct dep of docs

simshanith:
Stacked on #526 — the invariant only holds once that PR's devDeps land. Review #526 first.

## Why

#526 replaced twelve hardcoded `<pkg>#docs:api` edges in `docs/turbo.json` with a plain `^docs:api`, backed by declaring the four producers as devDeps of `docs`. `^` walks **direct deps only**, so those declarations are what make the fan-out reach all four.

That is already enforced by the build, loudly. `docs/api` is gitignored (`docs/.gitignore:3`) — 106 of its 107 files are generated — so CI always starts from an empty directory. Drop a producer's dep and you get:

```
EXIT=1
[UNRESOLVED_IMPORT] Could not resolve '../api/lit-ui-router-mobx/typedoc-sidebar.json'
```

What that does *not* do is explain itself. `package.json` takes no comments, so nothing in `docs/package.json` records why a site that never imports `lit-ui-router` devDepends on it. A knip/depcheck pass — or a person tidying — reads those three entries as unused. Locally, on a warm tree with stale `docs/api` contents, removing one still passes.

## What

`check-docs-api-deps.ts`: enumerate workspace members via `@tools/shared/workspace.ts`, take those with a `docs:api` script, assert each is in `docs`'s `dependencies ∪ devDependencies`.

Wired exactly like `//#check:patches` — a root script plus a `//#` turbo task on the root `lint` task's `with` list. No new CI wiring, no new package.

## Verification

Both arms, plus the wiring:

| | |
|---|---|
| passing arm | `4 docs:api producers are all direct deps of docs`, exit 0 |
| failing arm | one line per missing producer, plus the exact `"name": "workspace:*"` entries to add; exit 1 |
| `with` fan-out | `--dry=json` shows `//#check:docs-api-deps` pulled into any package's `lint` |
| failure propagation | broke `docs/package.json`, ran `turbo run lint` → turbo exit 1, `Failed: //#check:docs-api-deps` |
| empty-producer guard | fails rather than passing vacuously if the `docs:api` task ever disappears |

Type-aware oxlint caught that the new root `.ts` needed adding to `tsconfig.json`'s `include`, or `console`/`process` resolve as `error`-typed and the `no-unsafe-*` rules fire — same trap as the `.mjs` override. Fixed here, so `lint:root` is clean.

## Not done, deliberately

Making the import resolve through the producer package (`import ... from 'lit-ui-router-mobx/…'`) would turn this convention into a real, unprunable dependency. It is not available: `typedoc-vitepress-theme` hardcodes the sidebar at `path.resolve(outDir, 'typedoc-sidebar.json')`, and derives `basePath` from `relative(docsRoot, outDir)` — that prefix is baked into every link inside the JSON. The file is only *valid* under `docs/`. Relocating it would mean moving the markdown too and copying it back before VitePress scans routes.


docs/turbo.json
simshanith, line 35:
```suggestion
```

Issue #553, opened 2026-08-12
a11y: uiSrefActive lacks aria-current; no focus management on ui-view transition


simshanith:
Two accessibility gaps in `lit-ui-router`, found while surveying what a11y test surface the repo has. Both are in the shipped library rather than the sample apps, and neither is findable by a static axe scan of a page at rest — they only show up when you navigate and then assert.

Filing them together because they share a root question: **how much of client-side-navigation accessibility is the router's job?**

## 1. `uiSrefActive` conveys active state visually only

`packages/lit-ui-router/src/ui-sref-active.ts:241-255` toggles a CSS class and does nothing else. There is no `aria-current` anywhere in the repo (`grep -rn "aria-" packages/*/src` returns zero functional hits).

So every nav built with the documented pattern — including the one in the README and in `examples/helloworld/src/main.ts:32-38` — marks the current page for sighted users and not for assistive technology.

**This one looks like table stakes rather than a nice-to-have.** Every mainstream router does it:

| Router | Behaviour |
| --- | --- |
| [React Router `NavLink`](https://reactrouter.com/api/components/NavLink) | sets `aria-current="page"` automatically when active |
| [Vue Router `RouterLink`](https://router.vuejs.org/api/interfaces/routerlinkprops) | `ariaCurrentValue` prop, **defaults to `"page"`**, applied when exact-active |
| [Angular `RouterLinkActive`](https://angular.dev/api/router/RouterLinkActive) | `ariaCurrentWhenActive` input — **opt-in**, added in [angular/angular#45167](https://github.com/angular/angular/pull/45167) for [angular/angular#35051](https://github.com/angular/angular/issues/35051) |

Note the two API shapes: Vue defaults it on and lets you change the value; Angular makes it opt-in. `uiSrefActive` already distinguishes active from exact-active, which maps onto `aria-current`'s value set (`page`, `step`, `location`, `true`) — so the design question is which of those two shapes to copy, not whether the feature is wanted.

## 2. No focus management, announcement, or title update on transition

`<ui-view>` renders into light DOM (`packages/lit-ui-router/src/ui-view.ts:90-92`) and swaps `this.component(...)` output on transition (`ui-view.ts:361-388`). Nothing moves focus, nothing writes to a live region, and nothing sets `document.title` — `document.title` appears nowhere in `packages/*/src` or `apps/*/src`.

[...]

PR #556, opened 2026-08-12
feat(lit-ui-router): set aria-current on active ui-sref links

simshanith:
Closes the first half of #553 — `aria-current` in `uiSrefActive`. The second half of that issue (focus management / route announcement on `<ui-view>` transition) is deliberately **not** touched here; it is a contested design question and needs its own decision.

## The gap

`uiSrefActive` toggled a CSS class and did nothing else. `grep -rn "aria-" packages/*/src` returned zero functional hits. So every nav built with the documented pattern — the README's, the tutorial's, `examples/helloworld` — marked the current page for sighted users and not for assistive technology.

For what it's worth, `aria-current` appears zero times across `ui-router/react`, `ui-router/angular`, and `ui-router/core`, so this binding is the first in the UI-Router family to ship it. That is weak evidence of anything except that the family has not shipped much since the pattern became standard (core's last substantive releases: 6.1.1 Jul 2024, 6.1.2 Dec 2025). React Router documents `<NavLink>`'s entire purpose as providing exactly this context to assistive tech.

## What landed

```ts
uiSrefActive({ activeClasses: ['active'] })
// on a link, while the exact state is active: aria-current="page"
// otherwise:                                  attribute removed

uiSrefActive({ activeClasses: ['active'], ariaCurrentValue: 'step' })  // another token
uiSrefActive({ activeClasses: ['active'], ariaCurrentValue: false })   // hands off entirely
```

`ariaCurrentValue?: AriaCurrentValue | false`, where `AriaCurrentValue` is `'page' | 'step' | 'location' | 'date' | 'time' | 'true'`. It sits alongside `activeClasses`/`exactClasses` in the existing `UiSrefActiveParams` options object — no new argument position.

Three decisions, each of which could have gone the other way.

### 1. Default on, not opt in — Vue's shape, not Angular's

Vue Router's `RouterLink` defaults `ariaCurrentValue` to `"page"`; React Router's `NavLink` sets it unconditionally; Angular's `RouterLinkActive` requires you to ask for it. Angular's opt-in is a back-compat artifact of retrofitting a directive that had already shipped ([angular/angular#45167](https://github.com/angular/angular/pull/45167)), and the predictable result is that most Angular apps still do not set it. Default-off silently keeps the accessibility bug for everyone who does not read the docs, and adding `aria-current="page"` to a link that is already visibly highlighted is close to zero risk.

### 2. Exact match only

`aria-current="page"` means *this is the current page*. An ancestor state being active means the link points at a section **containing** the current page. A nav in which three ancestor links all claim `aria-current="page"` is worse for a screen reader user than a nav with none, so the attribute is applied on the `exact` branch, next to `exactClasses`.

Worth being precise about the precedent here, because it is split rather than unanimous: Vue applies on exact-active. React Router's `NavLink` applies on non-exact active unless you pass `end` — so at `/tasks/123`, the `/tasks` link is `aria-current="page"` too. Angular likewise follows `routerLinkActiveOptions`, which is non-exact by default. This PR follows Vue, and treats the React/Angular defaults as the wart rather than the model.

The consequence: `aria-current="location"` for an *ancestor* link (which is what that token is for) is not expressible. That would need a second key — something like `ariaCurrentActiveValue` — and it is a deliberate non-goal here rather than an oversight.

### 3. Default on for **links** only

This is the one departure from a straight Vue port, and the code forced it. `RouterLink` and `NavLink` are link components; there is no case where they land on a non-link. `uiSrefActive` is a directive that attaches to any element, and **the repo's own samples put it on wrappers**: `NavHeader.ts` on `<li>`, `MessageTable.ts` on `<tr>`, `MyMessages.ts` on `<li>`. Defaulting on unconditionally would have stamped `aria-current="page"` onto `<li>` and `<tr>` elements across the sample apps — valid ARIA, almost never what the author meant.

So the default applies to `<a>`, `<area>`, and `[role="link"]`. Any other element opts in by passing `ariaCurrentValue` explicitly. Wrappers can keep the classes while the inner link carries the attribute, which is the correct markup anyway and is covered by a test:

```html
<li ${uiSrefActive({ activeClasses: ['active'] })}>
  <a ${uiSref('users')} ${uiSrefActive({})}>Users</a>
</li>
```

`ariaCurrentValue: false` is a true no-op — the directive neither sets nor removes the attribute, leaving an author-managed `aria-current` alone. When enabled, removal on inactive matches how the sibling `uiSref` directive handles a null `href` (`ui-sref.ts:132-136`); it never writes `aria-current="false"`.

## In-repo usages

- **`examples/helloworld`, `hellogalaxy`, `hellosolarsystem`, `docs/tutorial/*`, README nav — no change needed.** They are plain `<a>` elements, so they get `aria-current="page"` for free. That is the point of the default, and it is the case the issue named.
- **`ContactList.ts` — no change needed.** `uiSrefActive` is already on the `<a>`.
- **`MyMessages.ts` — changed.** Folder list kept `selected` on the `<li>` and gained `${uiSrefActive({})}` on the inner leaf link, demoing the composed pattern.
- **`MessageTable.ts` — changed.** The `<tr>` opts in with `ariaCurrentValue: 'true'`, the idiomatic token for a selected row, demoing explicit opt-in on a non-link.
- **`NavHeader.ts` — left alone, deliberately.** Its tabs point at section states (`mymessages`, `contacts`, `prefs`) while the user is virtually always on a child state, so an exact-gated `aria-current` would essentially never fire. Adding markup that never activates would be noise. This is the case that wants ancestor `location` support — see the non-goal above.

## Docs

- Doc comments on `uiSrefActive` and `UiSrefActiveParams` (these drive the typedoc-generated API reference page).
- `docs/api/index.md` — new "Accessible active links" section covering the default, the exact-match rationale, the non-link opt-in, and disabling.
- `docs/tutorial/helloworld.md` and `README.md` feature list.

## Tests

Nine new specs in `ui-sref-active.spec.ts` (165 → 174 across the package): set when exact-active, **removed** when it goes inactive, not set on an active-but-not-exact ancestor, configured value, disabled no-op, non-link default-off, non-link explicit opt-in, `role="link"` default-on, and the wrapper-plus-inner-link composition.

```
lit-ui-router:test: ✓ happy-dom src/specs/ui-sref-active.spec.ts (52 tests) 4432ms
lit-ui-router:test:  Test Files  10 passed (10)
lit-ui-router:test:       Tests  174 passed (174)

 Tasks:    22 successful, 22 total
```

`turbo run test typecheck lint --filter=lit-ui-router` green (includes `typecheck:lit2`, `lint:templates`); `turbo run typecheck lint` green across the three sample-app packages.

## Update: rebased onto `main`, and the link predicate is now shared

Rebased onto current `main`, which carries #588, #589 and #596. No conflicts — this PR only touches `ui-sref-active.ts`, and the merged work is all in `ui-sref.ts`.

That rebase brought `isNativeLink` into reach, and reusing it fixes a bug this PR would otherwise have shipped. The predicate here asked its own tag question:

```ts
const tagName = element.tagName;
return tagName === 'A' || tagName === 'AREA' || element.matches('[role~="link"]');
```

`tagName` is `'A'` for an HTML anchor but `'a'` for an **SVG** anchor, so an SVG link never got the default `aria-current`. `isNativeLink` reads `localName`, which is lowercase for both, and #589 wrote it that way deliberately. So:

```ts
const isLinkElement = (element: Element): boolean =>
  isNativeLink(element) || element.matches('[role~="link"]');
```

`isNativeLink` becomes an `@internal` export from `ui-sref.ts`, alongside `UiSrefElement` and `UI_SREF_TARGET_EVENT`, which this file already imports from there. The dependency stays one-way.

**The two predicates stay separate on purpose**, and each now carries a doc comment saying why, because the obvious next refactor is to merge them:

- `aria-current` is **role**-based — `<div role="link">` legitimately takes it
- `href` is **tag**-based — the same element must never take one

They overlap on `<a>`/`<area>` and nowhere else. Sharing the tag half keeps that overlap exact instead of letting two hand-written copies drift, which is precisely what had already happened.

One spec added for the SVG anchor, mutation-tested: restoring the `tagName` comparison fails that spec and only that spec. 210 passing across the package.

## Update: rebased onto `main` after #590, and the `<tr>` now shows the split

#590 merged, so this rebase had one real conflict — `MessageTable`'s row, which both PRs had opinions about. That is not an accident of text; it is the one element in the app where the two predicates deliberately disagree, and the resolution keeps both:

```ts
// the one element where the two predicates part company. no href: that
// is tag-based, and a <tr> has none. aria-current anyway: that is
// role-based, off by default here, and opted into explicitly because a
// selected row really is the current location
html`<tr
  ${uiSrefActive({ activeClasses: ['active'], ariaCurrentValue: 'true' })}
  ${uiSref('.message', { messageId: message._id }, { assignHref: 'auto' })}
>
```

`assignHref: 'auto'` from #590 withholds the `href`; `ariaCurrentValue: 'true'` from this PR grants the `aria-current`. Same element, opposite answers, both correct — which is the argument for two predicates made concrete rather than asserted in a doc comment.

**#601 does not affect this PR's surface.** It converts twelve controls from `<button>` to `<a>`, and an anchor is exactly what gets `aria-current` by default here — but none of those twelve carries `uiSrefActive`. Every `uiSrefActive` host in the app is a wrapper `<li>`, the `<tr>`, or an anchor that was already an anchor. Checked, not assumed.

223 specs green across the package (up from 210 — main's #590 specs came with the rebase).

## Summary by CodeRabbit

* **New Features**
  * Active navigation links now automatically receive appropriate `aria-current` attributes.
  * Configure exact-match and ancestor-match values, or disable automatic ARIA handling.
  * Accessibility support includes links, SVG anchors, and non-link elements.
  * Active message-table rows now expose `aria-current="true"` for improved accessibility.

* **Bug Fixes**
  * Improved attribute ownership and cleanup while preserving authored `aria-current` values.

* **Documentation**
  * Expanded API and tutorial guidance with configuration examples and accessibility details.


apps/sample-app-shared/src/app/mymessages/MessageTable.ts
```diff
@@ -84,10 +84,15 @@ export class MessageTable extends LitElement {
       [...messages].sort(orderBy(sort)),
       ({ _id }) => _id,
       (message) =>
-        // a <tr> is neither a link by tag nor by role, so it takes no href
-        // under 'auto' and no aria-current by default — only the row class
+        // the one element where the two predicates part company. no href: that
+        // is tag-based, and a <tr> has none. aria-current anyway: that is
+        // role-based, off by default here, and opted into explicitly because a
+        // selected row really is the current location
```
simshanith: disallow such historical context dependent comments
summarize why aria-current follows here, even naming the `location` expectation (ref source code that discriminates; don't re-explain its decision)

simshanith: Rewritten in `9f815b3`. Drops the PR-narrative framing, points at `isLinkElement` (`ui-sref-active.ts`) as the discriminator instead of re-deriving its role-vs-tag reasoning, and names the token a reader would expect:

```ts
// 'auto' keeps href off a <tr>, which is no anchor; aria-current is off
// by default on the same shape (isLinkElement, ui-sref-active.ts) and is
// opted into here because a selected row is the current item in a set —
// 'true', not the 'location' that marks a step in a flow
```

Both halves are load-bearing at the call site: `assignHref` still defaults to `true` in 1.x, so without the explicit `'auto'` the row would take an href.

packages/lit-ui-router/src/ui-sref-active.ts
```diff
@@ -254,9 +344,74 @@ export class UiSrefActiveDirective extends AsyncDirective {
...
+        ? { exact: ariaCurrentValue }
+        : ariaCurrentValue === false
+          ? { exact: false }
+          : ariaCurrentValue;
+
+    const resolved = this.exact
+      ? (values.exact ?? (isLinkElement(this.element!) && 'page'))
+      : this.active
+        ? (values.active ?? false)
+        : false;
```
simshanith: refactor to avoid nested ternaries

simshanith: Refactored in `5c5c93b`. Both chains are gone, split by what they do:

`toAriaCurrentValues` normalizes at module scope, next to `isLinkElement`. The three-way chain collapses to one check — anything that is not the object form is an `exact` value, so `'page'`, `false` and `undefined` all take the same branch instead of two nested tests distinguishing them.

`resolveAriaCurrent` is a private method with early returns, which also gives the exact-wins-over-`active` rule somewhere to live as a signature (`AriaCurrentValues -> AriaCurrentValue | false`) rather than as ternary nesting depth.

The old JSDoc made two separate claims, so it split too: the exact/active precedence went to the resolver, and the "resolving to a single value keeps *no opinion*, *explicitly off* and *not applicable* from each needing their own branch" rationale stayed with `applyAriaCurrent`, since that is the method it justifies. Behaviour unchanged; 225 specs green.

docs/api/index.md
```diff
@@ -73,7 +73,80 @@ router.start();
...
+
+  ```html
+  <li ${uiSrefActive({ activeClasses: ['active'] })}>
+    <!-- classes on the wrapper, aria-current on the link -->
+    <a ${uiSref('users')} ${uiSrefActive({})}>Users</a>
+  </li>
+
+  <!-- an explicit value opts a non-link element in -->
+  <tr ${uiSref('.message', { messageId })}
+      ${uiSrefActive({ activeClasses: ['active'], ariaCurrentValue: 'true' })}></tr>
```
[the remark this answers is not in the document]
simshanith: Valid, fixed in `5c5c93b` at both sites.

Worth being explicit about why, since it is not only a consistency nit: `assignHref` still defaults to `true` in 1.x (`ui-sref.ts:280`), so these examples as written would write an `href` to a `<tr>` **and** trip the dev warning at `ui-sref.ts:296` telling the reader to pass `{ assignHref: 'auto' }` — the option the example omitted. A reader copying it would hit the warning immediately.

Both now read `${uiSref('.message', { messageId }, { assignHref: 'auto' })}`, matching `MessageTable`, with the comment naming what the option buys.

Swept for the same shape elsewhere: the only remaining non-link `uiSref` call sites without the option are in `ui-sref.spec.ts`, where `<button ${uiSref('home')}>` is deliberate — those specs exercise the default-true warn path.

simshanith:
Missed this one on the first pass — I was pulling inline review comments through `pulls/{n}/comments`, and this nitpick lives in the review summary body, which that endpoint does not return. Noting it so the gap does not repeat.

Valid, and added in `d08a386`. The claim checks out: all three existing object-form specs (lines 665, 688, 712) pass `exact` explicitly, so nothing covered an object that supplies `active` and leans on the `exact` default.

The `5c5c93b` refactor makes it more worth pinning, not less. `toAriaCurrentValues` now passes the object form straight through on `typeof === 'object'`, and the default lives one layer away in `resolveAriaCurrent` as `values.exact ?? (isLinkElement(...) && 'page')`. That separation is the silent-drop path the comment describes — a normalizer that helpfully filled in `exact: false` would look reasonable in isolation and break the link default.

Verified the spec actually bites rather than just passing. Mutating the normalizer to `{ exact: ariaCurrentValue.exact ?? false, active: ariaCurrentValue.active }` fails **this spec alone**, 1 of 226 — the precise regression it is meant to catch, with no collateral coverage masking it. (Breaking the default outright in `resolveAriaCurrent` fails 9, which is why that mutation proves nothing about this test in particular.)

Took the proposed test close to verbatim; the only change is a comment naming what the assertion pins.

PR #557, opened 2026-08-12
build(eslint): adopt eslint-plugin-lit, -wc and -lit-a11y over src

simshanith:
Adopts `eslint-plugin-lit`, `eslint-plugin-wc` and `eslint-plugin-lit-a11y`
over `{packages,apps,examples}/*/src/**/*.ts`.

## Why

`//#lint:templates` (lit-analyzer, #543/#547) gates template *correctness* —
unknown tags, attributes, properties, events. Its documented blind spot is
lifecycle and reactivity semantics: delete a `@property` and the gate stays
green. These three plugins read the class AST and cover exactly that.

The ESLint lane already existed for manifests, and already ended with
`oxlint.buildFromOxlintConfigFile` commented as a no-op placeholder. That
layer now does real work and stays last.

## Rules and severities

| Scope | lit | wc | lit-a11y |
| --- | --- | --- | --- |
| `packages/*/src` | error | error | error |
| `apps/*`, `examples/*` | error | error | **warn** |
| specs / test fixtures | error | error | off |

- `lit`: `flat/recommended`, plus `lit/lifecycle-super` and
  `lit/no-legacy-imports`.
- `wc`: `flat/best-practice` (recommended plus
  `require-listener-teardown`, `no-child-traversal-in-connectedcallback`,
  `no-typos`, ...).
- `lit-a11y`: `recommended`.

`lit-a11y` reports rather than gates in `apps/` and `examples/` on purpose:
that markup is a faithful port of `ui-router/sample-app-react`, and gating
repo CI on a 2016 demo's markup would block unrelated PRs. 23 warnings
today, all there.

### Off, with reasons in config

- `wc/guard-super-call` — every element here descends from `LitElement`,
  which always implements the callbacks and documents the *unguarded*
  `super.connectedCallback()` as required. The rule's premise holds only for
  mixins over a bare `HTMLElement`. 5 hits, all false.
- `wc/no-constructor-params` — constructor injection (`new Component(props)`)
  is a documented `RoutedLitElement` API. 12 hits, all false.
- `lit/attribute-names`, `lit/no-native-attributes` — overlap lit-analyzer's
  `no-unknown-attribute` / `-property`, which resolve against real element
  types. Left to lit-analyzer.
- `wc/no-exports-with-element` — evaluated, not adopted: file organisation,
  not element semantics.

One justified suppression pair, in `apps/sample-app-shared/.../Dialog.ts`.

## Findings in `packages/`

None. The two `<ui-view>` / `<ui-router>` listener sites were checked by hand
(see below) and the rest of the library is clean under all three plugins.

The interesting result is *why* it was clean before this PR and would have
stayed falsely clean without one line of config — see `settings.wc` below.

## Plumbing

Three seams were needed. Each is commented at its site.

**`@tools/eslint-ts-parser`** — ESLint cannot parse `.ts` without a TS parser,
and typescript-eslint hard-errors on TypeScript 7 (no JS API). pnpm resolves a
peer from the *importer*, so neither `packageExtensions` (the lit-analyzer
trick) nor scoped `overrides` can redirect it — both were tried and confirmed
not to work. A one-file workspace package carrying
`typescript: catalog:typescript6-compat` is the hop that makes the peer land
on TS 6. Syntax-only parse: no project service, no type information.

**`eslint.oxlint-directives.ts`** — ESLint 10 hard-errors on an
`eslint-disable` directive naming a rule it cannot resolve
(`linter.js` `createDisableDirectives`), with no opt-out, and oxlint owns
those directives in src. Every namespaced rule `.oxlintrc.json` enables is
registered as a no-op so the directives parse; oxlint stays the only
enforcer. Derived from `.oxlintrc.json` so a new oxlint rule cannot trap a
future PR.

**`settings.wc.elementBaseClasses: ['LitElement']`** — `wc` infers element
base classes by `require.resolve('lit')` from cwd. The root has no `lit` dep,
so `LitElement` was not in the base list, so every element registering from a
sibling `*.register.ts` rather than a `@customElement` decorator — including
`<ui-view>` and `<ui-router>` — was invisible to the whole plugin. Without
this line the lane passes while checking almost nothing in `packages/`.

## Wiring

`//#lint:elements` joins the `lint` task's `with` fan-out, alongside
`//#lint:package-json` and `//#lint:templates`. No `dependsOn`: the lane is
syntax-only, so it stays off the build path. Cache under `.cache/`
(already gitignored) via `--cache --cache-strategy=content`.

Measured: **1.4s cold, 1.1s warm** over 112 files.

## `wc/require-listener-teardown` — verdict: benign, no change made

The lead was `ui-router.ts:94` (adds `uiRouterContextEventName` in
`connectedCallback`, no `disconnectedCallback` at all) and `ui-view.ts:191`
(`seekRouter()` adds it; `disconnectedCallback` removes only
`uiViewContextEventName`).

Both handlers are stable per-instance arrow class fields
(`private readonly onUiRouterContextEvent = (event) => ...`). Per DOM spec
`addEventListener` is idempotent for an identical (type, callback, capture)
triple, so re-connecting cannot stack duplicates:

```
3 connects, 1 dispatch, stable field reference  -> 1 invocation
3 connects, 1 dispatch, fresh inline arrow      -> 3 invocations
```

Nothing leaks either: the listener is on `this` and closes over `this`, so it
is collected with the element.

The rule does not fire on these sites regardless — it requires a string
literal event name (both use `this.constructor.uiRouterContextEventName`) and
`ui-view`'s add is in `seekRouter()`, outside `connectedCallback`. No
suppression was needed and none was added.

The remaining asymmetry (`ui-view` tears down one of its two context
listeners) is cosmetic. Deliberately **not** bundled into a linter-adoption
PR; `ui-sref-active.ts` and the specs are under concurrent edit.

## Summary by CodeRabbit

* **Chores**
  * Added dedicated linting for custom elements across packages, apps, and examples.
  * Expanded checks for Lit patterns, Web Components standards, and accessibility concerns.
  * Integrated custom-element linting into the standard project validation workflow.
  * Added supporting TypeScript tooling and configuration for consistent lint results.

* **Documentation**
  * Clarified intentional exceptions in dialog and event-listener test scenarios to improve linting transparency.
  * Added guidance for running and formatting custom-element lint tasks.


package.json
```diff
@@ -21,6 +21,7 @@
     "format:toml": "mise run format_toml",
     "lint": "turbo run lint",
     "lint:actionlint": "mise run lint_actionlint",
+    "lint:elements": "eslint --cache --cache-location .cache/eslint-elements --cache-strategy content --format tap \"{packages,apps,examples}/*/src/**/*.ts\"",
```
simshanith: tap format preferred for package.json to get accurate list of linted packages; less important here think stylish / default may suffice

simshanith: Dropped in `3ac5478` — `lint:elements` now takes eslint's default `stylish`, `lint:package-json` keeps `--format tap` where the per-file list is the point.

No wrapper needed, though: the arg order does work, as long as the `--` is left out.

```bash
pnpm lint:elements --format tap   # ✅ 114 ok lines
pnpm lint:elements -- --format tap  # ❌ No files matching the pattern "--format"
```

pnpm forwards unrecognized flags to the script verbatim, so the appended `--format` lands after the globs and the last one wins — eslint's optionator does not care that it follows positionals. It is only the explicit `--` that breaks it: pnpm passes the separator through too, and eslint reads everything after it as a file pattern. `turbo run //#lint:elements -- --format tap` also works, since turbo appends without the separator.

Documented in **TURBO.md** rather than at the call site (per the follow-up): a new *Choosing an ESLint Formatter* section covering why the two tasks default differently and the no-`--` override, plus `//#lint:elements` added to the root-task list, which had never listed it.

For scale: this lane covers 114 files and currently reports 35 warnings (31 `lit-a11y/anchor-is-valid`), so stylish prints ~35 lines where tap printed 114 `ok` plus the diagnostics.

simshanith:
## Follow-up: the `--` is the whole problem, and it is optional

The review thread went back and forth on whether the arg order blocks a formatter override, so here is the settled version — the ordering is fine, and `pnpm run … -- …` was the only thing breaking it.

| invocation | result |
| --- | --- |
| `pnpm lint:elements` | stylish, findings only — the new default |
| `pnpm lint:elements --format tap` | ✅ TAP, 114 files |
| `pnpm lint:elements -- --format tap` | ❌ `No files matching the pattern "--format"` |
| `turbo run //#lint:elements -- --format tap` | ✅ TAP |
| `eslint … "globs" --format tap` | ✅ TAP |

Two separate things were conflated:

**Position does not matter to eslint.** Optionator parses flags wherever they land, so a `--format` appended *after* the glob still wins, and the last one wins:

```console
$ eslint --format tap "packages/lit-ui-router/src/ui-sref.ts" --format stylish
$            # silent — stylish with zero findings, i.e. the appended flag won
```

**The literal `--` does matter.** pnpm forwards unrecognized flags to the script verbatim, but it forwards the separator too, and eslint treats everything after `--` as a file pattern:

```console
$ pnpm run lint:elements -- --format stylish
No files matching the pattern "--format" were found.
```

So the fix is subtractive on both sides: drop `--format tap` from the script, and drop the `--` from the invocation. No wrapper task — a mise wrapper would have bought only the separator handling that turbo already does natively (turbo appends passthrough args without re-emitting `--`), at the cost of another shell file on the shellcheck surface.

Landed in `3ac5478`, with the table's conclusions written up in **TURBO.md** rather than as a comment beside the script.

simshanith:
Filed #606 for the 35 warnings this lane now prints. Not a blocker here — the lane is warn-only and lands green — but the noise floor should not become permanent.

The split is 31 `anchor-is-valid` hits, every one of them `<a ${uiSref('state')}>`: the rule cannot see an element-part directive assigning `href` at runtime, so it fires on the library's most-copied correct pattern. That is a policy call (disable vs. 31 inline suppressions vs. upstream), not a sweep. The other 4 — a missing `alt`, two icon-only nav anchors with no accessible name, and a keyboard-dead `<span @click>` sort control — are genuine and worth fixing in the sample apps people copy from.

PR #575, opened 2026-08-15
fix(sample-app): collapse the nested New Contact link and button

simshanith:
Found while tabbing through `/app/contacts` on a branch preview: the **New Contact** control took two tab stops.

Rebased onto current `main`, which now carries #588, #589 and #596.

## The bug

`ContactList.ts` rendered an `<a>` wrapping a `<button>`:

```html
<a ${uiSref('.new')}>
  <button class="btn btn-primary">…</button>
</a>
```

Both elements are focusable, so keyboard users stopped twice on one control — first on the anchor (which `uiSref` had given `href="/contacts/new"`), then on the button, which has no destination at all. It is also invalid markup: `<a>`'s content model forbids interactive content.

Collapsed to a single `<button>` carrying `uiSref`, matching the button-shaped navigation used in `Home`, `Welcome`, `NotFound`, `EditContact` and `ContactView`. One tab stop, no visual change.

## Modifier-click on the resulting button — twice mis-stated, now resolved upstream

This section has been wrong twice, so it is worth stating what is actually true on `main` today.

**First version.** I recorded the lost cmd/middle-click as an inherent consequence of using a button. That was wrong: it was our click handler bailing on any modifier in order to defer to native browser behaviour, on an element that has none to defer to. The click was dropped entirely — a dead control.

**Second version.** I then described it as a live bug blocking this PR. That is now stale. #589 landed the fix, and `main` scopes the guard:

```ts
// scoped to links: these guards hand the click back to the browser, and a
// non-link has nothing to hand it back to
if (isNativeLink(element) && (isModifiedClick(event) || opensOffApp(element))) return;
```

`isNativeLink` is false for a `<button>`, so the guard short-circuits and the click falls through to `$state.go` + `preventDefault`. **Nothing about this PR is blocked, and the control is not dead.**

**What is still true.** The button is not equivalent to the anchor. Cmd-click now navigates *in the same tab*: the click is handled rather than dropped, but the "open in a new tab" intent is silently discarded, because there is no `href` for the browser to open. That is inherent to button-shaped navigation, and it is the honest cost of this change — the tab-stop fix is worth it, and it is also the strongest remaining argument for converting these controls to links.

## The larger question this exposes, deliberately not fixed here

`uiSref` on `main` still writes `href` unconditionally, and most of this app applies it to a bare `<button>`, rendering `<button href="/home">` — an attribute that does nothing there. Navigation works via the click handler, but those controls have no link semantics: announced as "button" rather than "link", no copy-link, no new-tab, and never eligible for `aria-current` (which is link-only by default).

#590 adds the `assignHref` option for this, defaulting to `true` in 1.x with a dev-mode warning naming `'auto'` as the fix. **This button becomes one of those warning sites.** The annotation is a separate sweep over every non-link `uiSref` in the sample apps rather than a one-off here; whichever of the two merges second picks up this call site.

Converting the sample app's buttons to real links remains the eventual answer, deferred deliberately: `.btn`/`.btn-primary` carry no styling in this app (only `button.btn i + span` for icon spacing), so anchors would need a small stylesheet to keep the button look. That is one decision across the whole app, not a fix to smuggle into this PR.

Also worth noting: #557's `eslint-plugin-lit-a11y` would **not** have caught the nesting — it ships no nested-interactive rule, and `anchor-is-valid` only validates the href value.

## Verification

`turbo run typecheck lint build --filter=sample-app-shared...` — including `lint:templates` (lit-analyzer). Rebased clean onto `main` with no conflicts.

### DIFF
--- a/apps/sample-app-shared/src/app/contacts/ContactList.ts
+++ b/apps/sample-app-shared/src/app/contacts/ContactList.ts
@@ -16,12 +16,13 @@ export class ContactList extends LitElement {
   }
 
   render() {
+    // One control, not an <a> wrapping a <button>: nesting them gave two tab
+    // stops and put interactive content inside a link, which the content model
+    // forbids. Matches the button-shaped navigation used elsewhere in the app.
     const newContact = html`
-      <a ${uiSref('.new')}>
-        <button class="btn btn-primary">
-          <i class="fa fa-pencil"></i><span>New Contact</span>
-        </button>
-      </a>
+      <button class="btn btn-primary" ${uiSref('.new')}>
+        <i class="fa fa-pencil"></i><span>New Contact</span>
+      </button>
     `;
     const contacts = repeat(
       this.contacts,


simshanith:
## Checked against the upstream sample apps

Cloned all three (`ui-router/sample-app-angularjs`, `-react`, `-angular`) and read the directive sources.

**The nesting is inherited verbatim, in all three.** Byte-for-byte the same control:

- AngularJS — `app/contacts/contactList.component.js:14` — `<a ui-sref=".new"><button class="btn btn-primary">`
- Angular — `src/app/contacts/contact-list.component.ts:15` — `<a uiSref=".new"><button class="btn btn-primary">`
- React — `src/contacts/components/ContactList.jsx:21` — `<UISref to={'.new'}><a><button className="btn btn-primary">`

So this is a decade-old inheritance, not a porting slip. This PR deliberately diverges from upstream.

**And React explains where the idiom came from.** `UISref` is `cloneElement(children, {href, onClick, className})` (`lib/components/UISref.js:59`) — it pushes `href` onto its single child, so the child *has* to be an anchor for the href to mean anything. The `<button>` inside is purely visual. The extra tab stop is the price of wanting button styling on a component that insists on an anchor. AngularJS and Angular then copied the markup without that constraint applying to them.

### Where each one puts `href`

| | `href` on `<button uiSref>` | click guard reads | modifier keys checked |
| --- | --- | --- | --- |
| AngularJS 1.x | **yes** — any non-`FORM` element | own element (`el.attr('target')`) | button, ctrl, meta, **shift, alt** |
| Angular 2+ | **no** — `AnchorUISref` selector is `a[uiSref]` | own element (`nativeElement.target`) | button, ctrl, meta |
| React | **yes** — cloned onto whatever the child is | **`e.target.attributes.target`** | button, ctrl, meta, **shift, alt** |
| lit-ui-router | **yes** | **`event.target`** | button, ctrl, meta |

Three corrections to what I wrote in the description:

1. **We are not the outlier on `href`.** AngularJS writes it to any non-form element and React clones it onto any child, exactly as we do. But the one framework that got a deliberate redesign — Angular 2+ — went the other way: `AnchorUISref` is a separate directive selected on `a[uiSref]` and injected `@Optional()`, so `<button uiSref>` simply never receives an href. The majority is two legacy designs; the considered one disagrees with them.

2. **Our `event.target` bug is shared with React**, which reads `e.target?.attributes?.target` (`lib/hooks/useSref.js:95`) and has the same failure: click the `<span>` inside a link and `target="_blank"` is not seen. AngularJS and Angular both read their *own* element. Two of four get it right, and they're the two that aren't React.

3. **Missing `shiftKey`/`altKey` is ours and Angular's**; AngularJS and React both check all four. Shift-click opening a new window is real browser behaviour, so this is a genuine gap regardless of the split.

### The design this points at

`uiSref` writing `href` to any element and `uiSrefActive` writing `aria-current` only to links (#556) are inconsistent rules about the same question. Angular's answer makes them agree: **href on links, `aria-current` on links** — one notion of "this element is a link", applied by both directives. Under that rule the `<button>` this PR keeps would carry no inert `href`, and the button-shaped navigation throughout the app would be honestly marked as *not* links, which is what it is.

That is a library change, not a sample-app one, and it belongs behind its own issue.

simshanith:
## Correction: AngularJS is the origin, not React

The previous comment claimed React's `cloneElement` constraint explains where the `<a><button>` idiom came from. That is backwards — the chronology says otherwise.

Traced with `git log -S` across all three repos. The pattern is introduced in **AngularJS, `eb031c6`, 2016-02-05** (*"feat(contacts): Allow new contacts to be created … closes #24"*), fully formed:

```html
<li>
  <!-- This link is a relative ui-sref to the contacts.new state. -->
  <a ui-sref=".new">
    <button class="btn btn-primary">
      <i class="fa fa-pencil"></i><span>New Contact</span>
    </button>
  </a>
</li>
```

That comment line has survived every port since, this repo included.

| | repo first commit | pattern appears |
| --- | --- | --- |
| AngularJS | 2015-10-18 | **2016-02-05** (`eb031c6`) |
| React | 2016-08-02 | 2016-08-02 (first commit, imported wholesale) |
| Angular | 2017-02-03 | 2017-02-04 |

**Why this matters.** In AngularJS `<button ui-sref>` works fine and receives an `href` — there was no technical reason to wrap it, and Bootstrap 3 explicitly supports `.btn` on `<a>`, so the correct idiom (`<a class="btn btn-primary" ui-sref=".new">`) was available and documented at the time. The nesting was simply a slip.

React inherited it six months later, and only *there* did it become load-bearing: `UISref` clones `href` onto its single child, so an anchor child is genuinely required. A 2016 styling slip that a later port retroactively turned into a structural requirement — and that every subsequent port, including this one, copied without re-examining.

Nothing in the fix changes. The reasoning behind it does: this is not a constraint anyone designed, in any of the four.

simshanith:
Filed the click-handler findings as #576 — `event.target` → `currentTarget`, missing `shiftKey`/`altKey`/`defaultPrevented`, and a token-aware `rel` check, with the nine-router survey as justification. Independent of this PR; nothing here blocks on it.

simshanith:
Superseded by #601, which converts this control to `<a class="btn btn-primary" ${uiSref('.new')}>` as part of making the sample apps' navigation controls real links.

That fixes the same two things this PR does — one tab stop instead of two, and no interactive content inside a link — and it reaches the state this PR's rewritten body argued for, where cmd-click actually opens a new tab instead of navigating in place.

Nothing here is lost. Closing is yours to make; I have left it open.

PR #595, opened 2026-08-16
docs(release): add the break-glass local publish runbook

simshanith:
CI is the blessed publish path and stays that way. But a pipeline outage should not be the moment anyone works out the fallback from first principles, so `RELEASE.md` now carries an **Emergency Local Publish (break glass)** section, sibling to the existing first-publish runbook.

It opens with what a workstation publish forfeits, each row paired with what has to replace it by hand:

| CI provides                                         | Locally you must                                          |
| --------------------------------------------------- | ---------------------------------------------------------- |
| A fresh checkout — `dist` starts empty              | Clean `dist` yourself; nothing else prunes it             |
| OIDC trusted publishing — no long-lived token       | Use a real npm token, then revoke it                      |
| Build provenance attestation                        | Accept its absence — the release ships unattested         |
| Cold-bake reconciliation against CI's verified pack | Accept a single unwitnessed ledger                        |

Then a seven-step procedure, and a pointer from the Overview so the section is findable before it is needed.

### Why the first row is the one that bites

Every publishable package declares `files: ["dist/**", ...]`, and **no build task cleans `dist` first** — the emit tasks write over what is already there, and a turbo cache restore extracts over the directory rather than pruning it. Anything the build once emitted and no longer does sits there indefinitely and packs.

Found live while verifying `published-diff` under pnpm 12 (#566): `lit-ui-router-mobx` was carrying a month-old `dist/specs/test-utils.*` from before `tsconfig.src.json` excluded `src/specs`. `check:published-diff` reported it correctly as ship-affecting drift against the published `0.5.0` — those files are absent from the registry tarball and `files` would have shipped them. Every other output in that directory was minutes old; only the orphans were stale.

Worth noting what that means for reading the report: a local `SHIPS CHANGES` that CI calls clean is stale `dist`, not release debt. Check mtimes before believing it. The check was catching non-hermetic drift, which is not a class it was designed for — it detects it for free by comparing the working tree's pack against the registry rather than against another local build.

### Documented rather than fixed

The hazard is a non-hermetic workspace, and CI is already hermetic — fresh checkout, empty `dist`. Real exposure is bounded to a local publish, which is exactly the break-glass path this section governs.

The build-side fix is also weaker than it looks: cleaning `outDir` in the emit tasks would only cover cache misses, since turbo's restore does not prune. Narrowing `files` off `dist/**` to explicit globs would be airtight, but it buys hermeticity that CI has by construction. Neither seemed worth the churn against a documented procedure.

### The trap the procedure exists to prevent

There are two tarballs and publishing the wrong one is easy:

- `@tools/release#pack:all` → `.cache/pack/` — what `check:pack`, `check:exports`, and `check:published-diff` read. It can be a turbo-cache-restored artifact, and is deliberately kept **off** the publish path.
- `//tools/release:pack` → `.cache/publish/` — a cold re-pack from source through a restore-free staging copy. No turbo task declares this path. This is the only thing that may reach the registry.

My own first draft of step 6 published `.cache/pack/`'s tarball, which would have defeated the reason that separation exists. The section now names both and says which is which.

It also notes that `//tools/release:reconcile` still works locally when CI did manage to produce a verified pack for the commit — the one forfeited guarantee that is sometimes recoverable. In a full outage there is no witness and you proceed on the bake alone.

Docs only — no behaviour change.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -12,6 +12,10 @@ The release process is automated through GitHub Actions with protected environme
 4. **Tag** - A green main CI run automatically tags the release
 5. **Publish** - Publish to NPM and create GitHub Release
 
+Every stage runs in GitHub Actions. Publishing from a workstation is a
+break-glass fallback for when the pipeline itself is unavailable — see
+[Emergency Local Publish](#emergency-local-publish-break-glass).
+
 ## Tag and Branch Conventions
 
 | Pattern                        | Purpose                    | Example                        |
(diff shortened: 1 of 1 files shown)


simshanith:
not worth capturing at this point; can revisit when docs restructured

